PR #72, opened 2020-05-19
Merge in Acme branch

schmerl:


src/rosdiscover/interpreter/summary.py
```diff
@@ -8,12 +8,20 @@
 
 @attr.s(frozen=True, slots=True, auto_attribs=True)
 class NodeSummary:
+
```
ChrisTimperley, line 11: Last nitpick: Can we get rid of this space?

schmerl: Done

PR #80, opened 2020-06-11
 Creates Arch.yml using rosnode list

RomeInRuins:
Changes to be committed:
	new file:   src/rosdiscover/Issue74.py

### DIFF
--- a/src/rosdiscover/Issue74.py
+++ b/src/rosdiscover/Issue74.py
@@ -1,0 +1,102 @@
+import time
+import collections
+import sys
+
+sys.path.append("/interpreter")
+from summary import NodeSummary
+
+
+from loguru import logger
+
+import roswire
+
+f = open("Arch.yml", "w")
+
+# enable logging
+logger.enable('roswire')
+
+
+# first, we create a new ROSWire session
+rsw = roswire.ROSWire()
+
+def toString(l): 
+    s = ''
+    l1 = list(l)
+    if len(l1) == 0: 
+        return '[]'
+    for i in l1:
+        (fmt, topic) = i
+        s = s + "\n  - format: " + fmt + "\n    name: " + topic + "\n" 
+    return s
+
+# we then launch a container for a given robot.
+# once the end of the given context is reached, including the event where an
+# exception occurs, the container will be automatically destroyed.
+image = 'therobotcooperative/turtlebot3'
+sources = ['/opt/ros/kinetic/setup.bash', '/ros_ws/devel/setup.bash']
+environment = {'TURTLEBOT3_MODEL': 'burger'}
+with rsw.launch(image, sources, environment=environment) as system:
+
+    # we then launch ROS inside the container
+    # again, once the end of the context is reached, ROS is killed.
+    with system.roscore() as ros:
+
+        # let's bring up the application
+        ros.roslaunch('turtlebot3_house.launch',
+                      package='turtlebot3_gazebo',
+                      args={'gui': 'false'})
+
+        # we need to wait for the nodes to finish their initialisation
+        time.sleep(30)
+        
+        node_names = list(ros.nodes)
+
+        # get an overview of the system state
+        state = ros.state
+
+        # get topic types
+        topic_to_type = ros.topic_to_type
+
+        # get service formats
+        
+        service_to_format = {}
+        for service_name in state.services:
+            service = ros.services[service_name]
+            service_to_format[service_name] = service.format.name
+        
+
+nodeSummaryDict = {}
+for n in node_names: #create the dictionary of NodeSummary objects
+    
+    p = []
+    for key in state.publishers:
+        pubs = state.publishers[key]
+        for i in pubs:
+            if i == n: 
+                p.append((topic_to_type[key], key))
+       
+    s = []
+
+    for key in state.subscribers:
+        subs = state.subscribers[key]
+        for i in subs:
+            if i == n:
+                s.append((topic_to_type[key], key))
+    
+    serv = []
+    
+    for key in service_to_format: 
+        path = n + '/'
+        if path in key:
+            serv.append((service_to_format[key], key))
+
+    obj = NodeSummary('', n, '/', '', '', False, '', False, p, s, [], [], [], serv, [], [])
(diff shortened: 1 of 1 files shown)


src/rosdiscover/Issue74.py
```diff
@@ -0,0 +1,102 @@
...
+        return '[]'
+    for i in l1:
+        (fmt, topic) = i
+        s = s + "\n  - format: " + fmt + "\n    name: " + topic + "\n" 
+    return s
+
+# we then launch a container for a given robot.
+# once the end of the given context is reached, including the event where an
+# exception occurs, the container will be automatically destroyed.
+image = 'therobotcooperative/turtlebot3'
```
schmerl, line 35: Rewrite to take these in as arguments - perhaps take in a config yaml the same way that rosdiscover does, and get it that way.

src/rosdiscover/Issue74.py
```diff
@@ -0,0 +1,102 @@
...
+sources = ['/opt/ros/kinetic/setup.bash', '/ros_ws/devel/setup.bash']
+environment = {'TURTLEBOT3_MODEL': 'burger'}
+with rsw.launch(image, sources, environment=environment) as system:
```
schmerl, line 38: let's refactor this into a function

src/rosdiscover/Issue74.py
```diff
@@ -0,0 +1,102 @@
...
+    # again, once the end of the context is reached, ROS is killed.
+    with system.roscore() as ros:
+
+        # let's bring up the application
+        ros.roslaunch('turtlebot3_house.launch',
+                      package='turtlebot3_gazebo',
+                      args={'gui': 'false'})
+
+        # we need to wait for the nodes to finish their initialisation
+        time.sleep(30)
```
schmerl, line 50: We should work out if there is a better way to do this. Sleeping for a fixed amount isn't very general. @ChrisTimperley is there a more principled way to do this generally in ROS? I seem to recall that it is complicated.

src/rosdiscover/Issue74.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # get service formats
+        
+        service_to_format = {}
+        for service_name in state.services:
+            service = ros.services[service_name]
+            service_to_format[service_name] = service.format.name
+        
+
+nodeSummaryDict = {}
```
schmerl, line 68: I think I would refactor this a separate function too. That takes a state and returns NodeSummaryDict.

I think also, for this, you could try using mypy typing like we do in the other parts of the code.

RomeInRuins:
I know that the checks will fail... I just needed to push whatever work I had so that I could pull it from another computer.

ChrisTimperley:
> I know that the checks will fail... I just needed to push whatever work I had so that I could pull it from another computer.

That's perfectly fine! :slightly_smiling_face: You should feel free to make more, smaller commits, rather than worrying about making fewer, larger ones. Also, don't forget to [add short but descriptive messages to your commits](https://www.freecodecamp.org/news/writing-good-commit-messages-a-practical-guide/).

src/rosdiscover/cli.py
```diff
@@ -83,6 +86,90 @@ def rosservice_list(args) -> None:
...
+        nodeSummaryDict.update({n: obj})
+    return nodeSummaryDict
+
+
+def dynamic_analysis(args):
+    print(args)
+    with open(args.config.name, 'r') as r:
+        data = yaml.safe_load(r)
+    r.close()
+    f = open("Arch.yml", "w")
```
schmerl, line 150: This should be passed in as a parameter, which defaults to `arch,yml`. Add it as `--output` to the arg parser.

src/rosdiscover/cli.py
```diff
@@ -83,6 +86,90 @@ def rosservice_list(args) -> None:
...
+
+def get_info(image, sources, environment, file, package):
+    rsw = roswire.ROSWire()
+    with rsw.launch(image, sources, environment=environment) as system:
+        with system.roscore() as ros:
+            ros.roslaunch(file,
+                          package=package,
+                          args={'gui': 'false'})
+
+            time.sleep(30)
```
schmerl, line 108: Add this as an optional parameter, but default it to 30.

src/rosdiscover/cli.py
```diff
@@ -83,6 +86,90 @@ def rosservice_list(args) -> None:
...
+    r.close()
+    f = open("Arch.yml", "w")
+    logger.enable('roswire')
+    image = data['image']
+    sources = data['sources']
+    if 'environment' in data:
+        environment = data['environment']
+    node_names, state, topic_to_type, service_to_format = get_info(image, sources, environment,
+                                                          args.launchfile, args.package)
+    nodeSummaryDict = create_dict(node_names, state, topic_to_type, service_to_format)
```
schmerl, line 158: I think the following code can be done more simply, like we discussed yesterday

src/rosdiscover/cli.py
ChrisTimperley: Also, no need to explicitly `r.close()` here. `with` automatically takes care of calling `open` and `close` for you :slightly_smiling_face: (`r` will be closed upon reaching the end of the `with` block`.)

ChrisTimperley: And, in general, use `with open(...) as f:` every time that you perform file operations.

src/rosdiscover/cli.py
```diff
@@ -81,6 +86,74 @@ def rosservice_list(args) -> None:
...
+    if 'environment' in data:
+        environment = data['environment']
+    if args.sleep:
+        sleep_time = args.sleep
+    else:
+        sleep_time = 30
+    node_names, state, topic_to_type, service_to_format = get_info(image, sources, environment,
+                                                                   args.launchfile, args.package, sleep_time)
+    node_summary_dict = create_dict(node_names, state, topic_to_type, service_to_format)
+    f.write(json.dumps(node_summary_dict, indent=4, separators=(". ", " = ")))
```
ChrisTimperley, line 153: The `separators` argument here should be dropped to avoid producing bad JSON.

PR #81, opened 2020-06-15
Arch.yml to Acme

mvkashyap:
modified:   src/rosdiscover/cli.py
	modified:   src/rosdiscover/interpreter/summary.py
	new file:   test.acme


src/rosdiscover/cli.py
```diff
@@ -12,6 +12,8 @@
 from .config import Config
 from .interpreter import Interpreter
```
ChrisTimperley, line 13: Can combine with import below into:
```
from .interpreter import Interpreter, NodeSummary
```

src/rosdiscover/cli.py
```diff
@@ -48,11 +50,37 @@ def launch(args) -> None:
 
 
 def generate_acme(args):
+    with open(args.il.name, 'r') as f: 
+        data = yaml.load(f, Loader=yaml.FullLoader) #generates dictionary from YAML
```
ChrisTimperley, line 54: Prefer `yaml.safe_load(f)` rather than using `yaml.FullLoader`.

ChrisTimperley: Also, it's fine to drop the comment at the end of the line.

src/rosdiscover/cli.py
```diff
@@ -48,11 +50,37 @@ def launch(args) -> None:
...
     """Generates an Acme description for a given roslaunch command."""
```
ChrisTimperley, line 55: This line is a docstring that describes the `generate_acme` method. To work as intended, it needs to appear directly underneath the `def generate_acme(args):` line.

src/rosdiscover/cli.py
```diff
@@ -48,11 +50,37 @@ def launch(args) -> None:
...
-    interpreter = _launch_config(args)
-    # nodes = [n.to_dict for n in interpreter.nodes]
 
-    acme_gen = AcmeGenerator(interpreter.nodes, args.acme, args.jar)
+    nodes = []
+    if args.il is not None:
```
ChrisTimperley, line 58: If `args.il` is `None`, the code should have thrown an exception when we try to `open` the file above. Maybe we want to add a check above?

src/rosdiscover/cli.py
schmerl: I think you also need to check whether the argument is used before you try to open the file associated with it.

src/rosdiscover/cli.py
```diff
@@ -48,11 +50,37 @@ def launch(args) -> None:
...
+        for d in data: #create the NodeSummary object from a dictionary, append to nodes list
```
schmerl, line 59: Pull this code out into a separate function.

src/rosdiscover/cli.py
```diff
@@ -130,6 +159,7 @@ def main() -> None:
     p.add_argument("--jar", type=str, help='Pointer to the Acme jar file', default=acme_jar_path)
 
     p.add_argument('config', type=argparse.FileType('r'), help=CONFIG_HELP)
+    p.add_argument('--il', type=argparse.FileType('r'))
```
schmerl, line 162: change this to --arch

test.acme
```diff
@@ -0,0 +1,236 @@
+import families/ROSFam.acme;
```
schmerl, line 1: I think this file can be removed from the repository.

RomeInRuins:
I fixed all the issues mentioned.

src/rosdiscover/cli.py
```diff
@@ -48,33 +46,34 @@ def launch(args) -> None:
...
+        same_fields = {'pubs' : [], 'subs' : [], 'uses' : [], 'provides' : [], 
+                        'action-servers' : [], 'action-clients' : []}
+        for i in same_fields:
+            for j in d[i]:
+                same_fields[i].append((j['name'], j['format']))
+        for i in d['reads']:
+            read.append((i['name'], i['dynamic']))
+        for i in d['writes']:
+            write.append(i)
+        obj = NodeSummary(d['name'], d['fullname'], d['namespace'],
```
ChrisTimperley, line 65: To avoid unexpected mishandling of arguments, use keyword arguments to build the node summary. (This is also resilient to any possible API changes.)
```
NodeSummary(name=d['name'], ...)
```

src/rosdiscover/cli.py
```diff
@@ -48,33 +46,34 @@ def launch(args) -> None:
     else:
         print(yaml.dump(output, default_flow_style=False))
 
+def nodes_from_yaml(args):
+    with open(args.arch.name, 'r') as f: 
+            data = yaml.safe_load(f) 
+    nodes = []  
+    for d in data: #create the NodeSummary object from a dictionary, append to nodes list
```
ChrisTimperley, line 53: No need for comment

src/rosdiscover/cli.py
```diff
@@ -48,33 +46,34 @@ def launch(args) -> None:
     else:
         print(yaml.dump(output, default_flow_style=False))
 
+def nodes_from_yaml(args):
+    with open(args.arch.name, 'r') as f: 
+            data = yaml.safe_load(f)
```
ChrisTimperley, line 51: Looks like indentation is off here. Should be 4 spaces.

PR #86, opened 2020-06-24
translator from LaunchConfiguration to YML

RomeInRuins:
Things that I need to update:
-Add a field for action servers and clients, or at least one for names
-Right now, there are some nodes that aren't discovered because the full file path doesn't seem to be available unless the generate_launch_description() function is called

### DIFF
--- a/src/rosdiscover/ros2_translator/arch.py
+++ b/src/rosdiscover/ros2_translator/arch.py
@@ -1,0 +1,57 @@
+import yaml 
+
+import os, sys
+
+
+from launch import LaunchDescription
+from launch.actions import IncludeLaunchDescription
+
+
+from launch_ros.actions import Node
+from launch.launch_description_sources import get_launch_description_from_python_launch_file
+
+
+
+description = get_launch_description_from_python_launch_file('robot.launch.py')
+nodes = description.entities
+
+
+def toYaml(nodes):
+    to_yaml = {}
+    no_name = 0
+    
+    for node in nodes:
+        if isinstance(node, IncludeLaunchDescription):
+            description1 = node._IncludeLaunchDescription__launch_description_source.try_get_launch_description_without_context()
+            if description1 == None:
+                continue
+            nodes1 = description1.entities
+            to_yaml.update(toYaml(nodes1))
+        
+        if isinstance(node, Node):
+            dic = {}
+            if node._Node__node_name != None:
+                name = node._Node__node_name
+                name_entry = name + 'name'
+                dic.update({name+'name': node._Node__node_name})
+            else:
+                name = 'unknown' + str(no_name)
+                name_entry = name
+                no_name+=1
+                dic.update({'name': name})
+            if node._Node__node_namespace != None:
+                namespace_entry = name + 'namespace'
+                dic.update({'namespace': node._Node__node_namespace})
+            if node._Node__node_executable != None:
+                exec_entry = name + 'executable'
+                dic.update({'executable': node._Node__node_executable})
+            if node._Node__package != None:
+                pack_entry = name + 'package'
+                dic.update({'package': node._Node__package})
+
+            to_yaml.update({name_entry: dic})
+    return to_yaml 
+
+with open('Arch.yml', 'w') as f:
+    print(yaml.dump(toYaml(nodes), f, default_flow_style=False))
+


src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
...
+    for node in nodes:
+        if isinstance(node, IncludeLaunchDescription):
+            description1 = node._IncludeLaunchDescription__launch_description_source.try_get_launch_description_without_context()
+            if description1 is None:
+                continue
+            nodes1 = description1.entities
+            to_yaml.update(toyaml(nodes1))
+        if isinstance(node, Node):
+            dic = {}
+            if node._Node__node_name is not None:
```
ChrisTimperley, line 22: In general, we shouldn't be accessing private attributes. This is really a hack that should be used in very few situations. I suspect that there are some publicly visible properties, methods, or attributes that we could use to get the same information?

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
...
+                name_entry = node._Node__node_name
+                dic.update({'name': node._Node__node_name})
```
ChrisTimperley, line 24:
```
dic['name'] = foo
```

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
...
+            else:
+                name = 'unknown' + str(no_name)
+                name_entry = name
```
ChrisTimperley, line 27: dead variable?

src/rosdiscover/ros2_translator/arch.py
RomeInRuins: Yeah it really is. When I looked, I couldn't find a method except for accessing the node name. I'll look again though.

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
...
+                no_name += 1
+                dic.update({'name': name})
+            if node._Node__node_namespace is not None:
```
ChrisTimperley, line 30:
```
if foo:
```
is a more Pythonic way to achieve the following:
```
if foo is not None
```

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
+import yaml
```
ChrisTimperley, line 1: I'm not 100% sure about the YAML implementation/package that's used by ROS2, but the `yaml` module isn't part of the Python standard library, so there's always a risk here that we will try to use code that isn't available inside the Docker container for a particular robot.

A more portable alternative here would be to use the `json` module, which is included in the Python standard library. Using `json` should have minimal impact on the rest of the script.

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
...
+nodes = description.entities
+
+
+def toyaml(nodes):
+    to_yaml = {}
+    no_name = 0
+    for node in nodes:
+        if isinstance(node, IncludeLaunchDescription):
+            description1 = node._IncludeLaunchDescription__launch_description_source.try_get_launch_description_without_context()
+            if description1 is None:
```
ChrisTimperley, line 16: This if-condition can't be reached. If `node` is `None`, then `isinstance(node, IncludeLaunchDescription)` will evaluate to `False`.

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
...
+description = get_launch_description_from_python_launch_file('robot.launch.py')
+nodes = description.entities
+
+
+def toyaml(nodes):
+    to_yaml = {}
+    no_name = 0
+    for node in nodes:
+        if isinstance(node, IncludeLaunchDescription):
+            description1 = node._IncludeLaunchDescription__launch_description_source.try_get_launch_description_without_context()
```
ChrisTimperley, line 15: What is `description1`? Prefer a descriptive variable name that communicates the purpose of this particular description.

RomeInRuins: Sorry about that. Description1 is the LaunchDescription accessed by node, if node is of class IncludeLaunchDescription. In other words, it's the LaunchDescription generated by the launch file included in IncludeLaunchDescription

src/rosdiscover/ros2_translator/arch.py
ChrisTimperley: `node.node_name` _should_ work here (https://github.com/ros2/launch_ros/blob/master/launch_ros/launch_ros/actions/node.py#L320)

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
...
+            if node._Node__node_name is not None:
+                name_entry = node._Node__node_name
+                dic.update({'name': node._Node__node_name})
+            else:
+                name = 'unknown' + str(no_name)
+                name_entry = name
+                no_name += 1
+                dic.update({'name': name})
+            if node._Node__node_namespace is not None:
+                dic.update({'namespace': node._Node__node_namespace})
```
ChrisTimperley, line 31: `node.expanded_node_namespace` _should_ do the trick here.

src/rosdiscover/ros2_translator/arch.py
ChrisTimperley: It looks like you're right. There are a few useful getter properties that give us access to some information, but other important parts are made private with `__`. Where we can't use the public API to get access to the information we need, this is OK, but we should maybe add a small comment in the code to justify the decision (and to remind us in the future that this script may break with changes to the implementation of `ros_launch` and `launch`).

ChrisTimperley:
Thanks for all of your hard work on this, Mehal! This is great. Let's tidy things up, add some command-line options, make it portable, and then we'll be good to go :-)

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
...
+            if node._Node__node_executable is not None:
+                dic.update({'executable': node._Node__node_executable})
+            if node._Node__package is not None:
+                dic.update({'package': node._Node__package})
+            to_yaml.update({name_entry: dic})
+    return to_yaml
+
+
+with open('Arch.yml', 'w') as f:
```
ChrisTimperley, line 40: For the sake of best practices, let's wrap this in the following:
```
if __name__ == '__main__':
    ...
```

The code above ensures that the contents of the if-block are only executed if the file to which they belong _is_ the Python script that's being executed (i.e., if you were to `import` this file in another file, then the code in the if-block wouldn't be executed.)

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,41 @@
+import yaml
+from launch.actions import IncludeLaunchDescription
+from launch_ros.actions import Node
+from launch.launch_description_sources import get_launch_description_from_python_launch_file
+
+description = get_launch_description_from_python_launch_file('robot.launch.py')
```
ChrisTimperley, line 6: Let's also move these two lines `description` and `nodes` to the body of the if-statement at the bottom of the file.

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,45 @@
...
+    no_name = 0
+    for node in nodes:
+        if isinstance(node, IncludeLaunchDescription):
+            description_from_node = node._IncludeLaunchDescription__launch_description_source.try_get_launch_description_without_context()
+            if description_from_node is None:
+                continue
+            nodes1 = description_from_node.entities
+            to_yaml.update(toyaml(nodes1))
+        if isinstance(node, Node):
+            '''
```
ChrisTimperley, line 18: Another Python tip:
```python
''''
these multi-line strings are used doc-strings, and typically only appear below a method or class definition. doc-strings are used to provide API information (i.e.,
what a class/method does) but not implementation details (i.e., how it does that thing).
""

# comments about the implementation of a method or a particular
# block of code should look like this. these comments are only seen
# by other developers editing this code and can't be seen by users
# of the method (they shouldn't have any knowledge of how the method
# actually works under the hood)
```

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,47 @@
...
+                dic['executable'] = node._Node__node_executable
+            if node._Node__package:
+                dic['package'] = node._Node__package
+            to_yaml[name_entry] = dic
+    return to_yaml
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Find nodes')
+    parser.add_argument('--launchfile', type=str)
```
ChrisTimperley, line 42: `add_argument('filename', type=str)` should be fine here (i.e., we use a positional argument rather than a keyword argument).

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,47 @@
...
+    args = parser.parse_args()
+    description = get_launch_description_from_python_launch_file(args.launchfile)
+    nodes = description.entities
+    with open('Arch.yml', 'w') as f:
```
ChrisTimperley, line 46: Potentially, we may want the ability to specify the output filename (maybe `--output-file` with `arch.yml` as a default?)

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,52 @@
...
+            if node._Node__package:
+                dic['package'] = node._Node__package
+            to_yaml[name_entry] = dic
+    return to_yaml
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Find nodes')
+    parser.add_argument('launchfile', type=str)
+    parser.add_argument('--output', type=str)
```
ChrisTimperley, line 43: `default="arch.yml"` should save a few lines :slightly_smiling_face:

ChrisTimperley: Also, I guess the default output should probably be `arch.json` now.

src/rosdiscover/ros2_translator/arch.py
```diff
@@ -0,0 +1,52 @@
...
+from launch_ros.actions import Node
+from launch.launch_description_sources import get_launch_description_from_python_launch_file
+
+
+def toyaml(nodes):
+    to_yaml = {}
+    no_name = 0
+    for node in nodes:
+        if isinstance(node, IncludeLaunchDescription):
+            description_from_node = node._IncludeLaunchDescription__launch_description_source.try_get_launch_description_without_context()
```
ChrisTimperley, line 13: The public API can be used to access `launch_description_source`:
```
node.launch_description_source
```

PR #97, opened 2020-07-23
Added subset of data structures for representing recoved architectural elements

ChrisTimperley:
This should cover most basic operations involving topics, parameters, services, and action servers.

### DIFF
--- a/src/rosdiscover/recover/core.py
+++ b/src/rosdiscover/recover/core.py
@@ -9,7 +9,8 @@ __all__ = ('PublisherDefinition',
            'ParameterRead',
            'ParameterSearch',
            'ParameterWrite',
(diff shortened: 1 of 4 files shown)


src/rosdiscover/recover/core.py
```diff
@@ -0,0 +1,257 @@
...
+        The name of the service.
+    """
+    source: str
+    location: sourcelocation.FileLocationRange
+    type_: str
+    service: str
+
+
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class ServiceProxyDefinition(RecoveredNodeModelElement):
```
schmerl, line 66: Is a server proxy the same as a service client? Perhaps "ServiceClient" would be a better name?

PR #99, opened 2020-07-27
Added workaround for Xacro deprecation warning (fixes #98)

### LINKED ISSUE #98: Failure during interpretation of Fetch launch config due to Xacro deprecation warning
```
  File "/home/chris/.pyenv/versions/3.7.5/lib/python3.7/xml/etree/ElementTree.py", line 1315, in XML
    parser.feed(text)
    │      │    └ '<?xml version="1.0" encoding="utf-8"?>\r\n<!-- =============================================================================...
    │      └ <method 'feed' of 'xml.etree.ElementTree.XMLParser' objects>
    └ <xml.etree.ElementTree.XMLParser object at 0x7fdfb91d3200>

  File "<string>", line None

xml.etree.ElementTree.ParseError: not well-formed (invalid token): line 938, column 0
```

This occurs due to the inclusion of the following warning inside the `stdout` from `xacro.py`:

```
xacro.py is deprecated; please use xacro instead
```

A simple fix is to just look for this particular string and remove it.

ChrisTimperley:
See #98

### DIFF
--- a/src/rosdiscover/models/spawn_model.py
+++ b/src/rosdiscover/models/spawn_model.py
@@ -43,6 +43,11 @@ def spawn_model(c):
     with open('urdf.xml', 'w') as f:
         f.write(urdf_contents)
 
+    # #94 workaround to deal with inclusion of warning when xacro.py is used
+    end_tag = '</robot>'
+    end_tag_starts_at = urdf_contents.rfind(end_tag)
+    end_tag_ends_at = end_tag_starts_at + len(end_tag)
+    urdf_contents = urdf_contents[:end_tag_ends_at]
     urdf_xml = ET.fromstring(urdf_contents)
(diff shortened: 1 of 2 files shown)


src/rosdiscover/models/spawn_model.py
```diff
@@ -43,6 +43,11 @@ def spawn_model(c):
     with open('urdf.xml', 'w') as f:
         f.write(urdf_contents)
 
+    # #94 workaround to deal with inclusion of warning when xacro.py is used
```
schmerl, line 46: I think this assumes that there is only one <robot> block in each file. Is that always the case?

ChrisTimperley: Yep.

PR #100, opened 2020-07-28
Import this branch which has gotten too big

schmerl:
This branch has gotten too big and complicated. I'd like to merge and use finer grained branches going forward.


src/rosdiscover/interpreter/context.py
```diff
@@ -191,12 +191,6 @@ def action_server(self, ns: str, fmt: str) -> None:
         ns = self.resolve(ns)
         self._action_servers.add((ns, fmt))
 
-        self.sub(f'{ns}/goal', f'{fmt}Goal')
```
ChrisTimperley, line 194: We shouldn't get rid of these. We _may_ want to add a new tag to these `implicit` topics that are created by action servers and services. If so, we should create a new issue.

schmerl: Architecturally, I'm not sure that that is true. My thinking is that we either have an action port and none of the other ports, or the other ports and not an action port. In Acme, the architecture connector for actions has roles for clients to keep track of progress and cancel and such and so anyone that subscribes or publishes those does so architecturally through the connector.

schmerl: In any case, I've added them back in and will raise an issue once this branch is merged.

ChrisTimperley: It sounds like Acme is influencing design choices in components that are supposed to be separate. (This code and the `interpreter` stuff should be unaware of the existence of the `acme` module.) If we drop these, we'll be producing an incomplete picture of the architecture. It sounds like the issue lies not here, but in the interaction between Acme and `NodeSummary`. I strongly suspect that adding `implicit=True` will solve this issue.

src/rosdiscover/interpreter/context.py
```diff
@@ -125,7 +125,7 @@ def use(self, service: str, fmt: str) -> None:
         service_name_full = self.resolve(service)
         self._uses.add((service_name_full, fmt))
 
-    def sub(self, topic_name: str, fmt: str) -> None:
+    def sub(self, topic_name: str, fmt: str, implicit: bool = True) -> None:
```
ChrisTimperley, line 128: `implicit` should probably be `False` by default

src/rosdiscover/interpreter/context.py
```diff
@@ -29,8 +29,8 @@ class NodeContext:
     _placeholder: bool = attr.ib(default=False, repr=False)
     _uses: Set[Tuple[str, str]] = attr.ib(factory=set, repr=False)
     _provides: Set[Tuple[str, str]] = attr.ib(factory=set, repr=False)
-    _subs: Set[Tuple[str, str]] = attr.ib(factory=set, repr=False)
-    _pubs: Set[Tuple[str, str]] = attr.ib(factory=set, repr=False)
+    _subs: Set[Tuple[str, str, bool]] = attr.ib(factory=set, repr=False)
```
ChrisTimperley, line 32: This is getting very stinky and dangerous. (`Tuple[str, str]` already wasn't great.) Every thing that interacts with `_subs` now needs to be changed and it's now easy to confuse the order of the items in the tuple and create sneaky new errors. The proper fix here is to introduce a new `Topic` data class with `name`, `type_`, and `implicit` parameters.

schmerl: Yeah. I've undone all this in this branch and I think you can now merge.

PR #102, opened 2020-07-28
Implicit topics that can be used to indicate action topics (fixes #101)

### LINKED ISSUE #101: Topics generated by actions should be indicated somehow
Currently, when actions are created they also generate the topics that are implied by the actions. When processing the architecture, this causes both actions and topics to be created in a node. Only the action is needed.

schmerl:
See #101


src/rosdiscover/interpreter/topic.py
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+__all__ = ('Topic',)
```
ChrisTimperley, line 2: Missing whitespace below this line

src/rosdiscover/interpreter/topic.py
```diff
@@ -0,0 +1,15 @@
...
+from typing import Any, Dict
```
ChrisTimperley, line 3: Missing whitespace

src/rosdiscover/interpreter/topic.py
```diff
@@ -0,0 +1,15 @@
...
+import attr
+
+
+@attr.s(frozen=True, slots=True)
```
ChrisTimperley, line 7: This is great. If you want to make it even tidier, you can add `auto_attribs=True` to the `attr.s` call and drop ` = attr.ib()` from each attribute. (See: https://github.com/ChrisTimperley/rosdiscover/blob/master/src/rosdiscover/interpreter/summary.py#L9-L12)

schmerl:
I think this can be merged now.

Issue #104, opened 2020-09-16
Add '_app' attribute to context


ChrisTimperley:
This will allow us, for example, to look at the ROS version that's being used.

PR #107, opened 2020-09-18
Updated to allow config and app (and eventually the ros version)

schmerl:
Fix for #104

### DIFF
--- a/src/rosdiscover/cli.py
+++ b/src/rosdiscover/cli.py
@@ -26,7 +26,7 @@ def _launch(config: Config) -> SystemSummary:
                                ) as interpreter:
         for fn_launch in config.launches:
             logger.info(f"simulating launch [{fn_launch}]")
-            interpreter.launch(fn_launch)
+            interpreter.launch(fn_launch, config)
         return interpreter.summarise()
 
 
--- a/src/rosdiscover/interpreter/interpreter.py
+++ b/src/rosdiscover/interpreter/interpreter.py
@@ -10,6 +10,7 @@ from .context import NodeContext
 from .model import Model
 from .summary import SystemSummary
 from .parameter import ParameterServer
+from ..config import Config
 
 
 class Interpreter:
@@ -43,7 +44,7 @@ class Interpreter:
         node_to_summary = {s.fullname: s for s in node_summaries}
         return SystemSummary(node_to_summary)
 
-    def launch(self, filename: str) -> None:
+    def launch(self, filename: str, configuration: Config) -> None:
         """Simulates the effects of `roslaunch` using a given launch file."""
         # NOTE this method also supports command-line arguments
         reader = LaunchFileReader(shell=self._app.shell,
@@ -68,7 +69,8 @@ class Interpreter:
                            namespace=node.namespace,  # FIXME
                            launch_filename=node.filename,
                            remappings=remappings,
-                           args=args)
+                           args=args,
+                           config=configuration)
             # FIXME this is waaay too permissive
             except Exception:
                 logger.exception(f"failed to launch node: {node.name}")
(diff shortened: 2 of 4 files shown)


src/rosdiscover/cli.py
```diff
@@ -26,7 +26,7 @@ def _launch(config: Config) -> SystemSummary:
                                ) as interpreter:
```
ChrisTimperley, line 26: We could improve this by replacing `Interpreter.for_image` with `Interpreter.for_config`. (Alternatively, we could be more OO-friendly by adding an `interpreter()` method to `Config`, that produces an interpreter for a given configuration.)

src/rosdiscover/cli.py
```diff
@@ -26,7 +26,7 @@ def _launch(config: Config) -> SystemSummary:
...
         for fn_launch in config.launches:
             logger.info(f"simulating launch [{fn_launch}]")
-            interpreter.launch(fn_launch)
+            interpreter.launch(fn_launch, config)
```
ChrisTimperley, line 29: If we make the changes above, we wouldn't need to pass `config` here.

src/rosdiscover/interpreter/interpreter.py
```diff
@@ -68,7 +69,8 @@ def launch(self, filename: str) -> None:
                            namespace=node.namespace,  # FIXME
                            launch_filename=node.filename,
                            remappings=remappings,
-                           args=args)
+                           args=args,
+                           config=configuration)
```
ChrisTimperley, line 73: We could drop the `config` parameter if `configuration` becomes a property of `Interpreter`.

src/rosdiscover/interpreter/interpreter.py
```diff
@@ -90,6 +92,7 @@ def _load_nodelet(self,
                       namespace: str,
                       launch_filename: str,
                       remappings: Dict[str, str],
+                      config: Config,
```
ChrisTimperley, line 95: See above

src/rosdiscover/interpreter/interpreter.py
```diff
@@ -132,7 +135,8 @@ def _load_nodelet(self,
                           namespace=namespace,
                           launch_filename=launch_filename,
                           remappings=remappings,
-                          args='')
+                          args='',
+                          config=config)
```
ChrisTimperley, line 139: See above

src/rosdiscover/interpreter/interpreter.py
```diff
@@ -141,7 +145,8 @@ def _load(self,
               namespace: str,
               launch_filename: str,
               remappings: Dict[str, str],
-              args: str
+              args: str,
+              config: Config
```
ChrisTimperley, line 149: See above

src/rosdiscover/interpreter/interpreter.py
```diff
@@ -182,7 +187,8 @@ def _load(self,
                                           name=name,
                                           namespace=namespace,
                                           launch_filename=launch_filename,
-                                          remappings=remappings)
+                                          remappings=remappings,
+                                          config=config)
```
ChrisTimperley, line 191: See above

src/rosdiscover/interpreter/interpreter.py
```diff
@@ -192,7 +198,8 @@ def _load(self,
                                           namespace=namespace,
                                           launch_filename=launch_filename,
                                           remappings=remappings,
-                                          manager=mgr)
+                                          manager=mgr,
+                                          config=config)
```
ChrisTimperley, line 202: See above

ChrisTimperley (changes requested):
This looks good, but there's a couple of places where we can clean up pre-existing design limitations to improve quality generally.

PR #112, opened 2020-09-28
Updated code to only produce implicit pubs/sub/services for particular ROS versions (fixes #105)

### LINKED ISSUE #105: Feature: Do not create implicit pubs/subs/services for actions in certain versions of ROS2
The easiest way to do this (although not the cleanest in an OOP sense) is to add an if-condition to the implementation of the `action_client` and `action_server` methods that checks the ROS version. The alternative, OO approach, is to have behavioral subclasses of `Context` that behave appropriately for different versions of ROS.

Depends on:
* [ ] https://github.com/ChrisTimperley/rosdiscover/issues/104
* [ ] https://github.com/ChrisTimperley/rosdiscover/issues/106
* [ ] https://github.com/ChrisTimperley/roswire/issues/398

schmerl:
Fixes #105.


src/rosdiscover/interpreter/context.py
```diff
@@ -199,7 +199,7 @@ def action_server(self, ns: str, fmt: str) -> None:
                      f"[{ns}] with format [{fmt}]")
         ns = self.resolve(ns)
         self._action_servers.add(Action(name=ns, format=fmt))
-        if self.app.description.distribution.ros == ROSVersion.ROS1:
+        if self.actions_have_topics():
```
ChrisTimperley, line 202: Maybe turn `actions_have_topics` into a `@property`?

ChrisTimperley: Actually, ignore this comment.

src/rosdiscover/interpreter/context.py
```diff
@@ -234,6 +234,11 @@ def action_client(self, ns: str, fmt: str) -> None:
             self.sub(f'{ns}/feedback', f'{fmt}Feedback', implicit=True)
             self.sub(f'{ns}/result', f'{fmt}Result', implicit=True)
 
+    def actions_have_topics(self):
+        distribution = self.app.description.distribution
+        return (distribution.ros == ROSVersion.ROS1) or (
```
ChrisTimperley, line 239: Could unpack:
```
if distro ... :
  return True
return distro ...
```

schmerl: Done.

src/rosdiscover/interpreter/context.py
```diff
@@ -234,6 +234,13 @@ def action_client(self, ns: str, fmt: str) -> None:
             self.sub(f'{ns}/feedback', f'{fmt}Feedback', implicit=True)
             self.sub(f'{ns}/result', f'{fmt}Result', implicit=True)
 
+    def actions_have_topics(self):
+        distribution = self.app.description.distribution
+        if distribution == ROSVersion.ROS1:
+            return True
+        else:
+            return distribution.ros == ROSVersion.ROS2 and distribution.name < ROSDistribution.FOXY
```
ChrisTimperley, line 242: These changes are good, but I don't think that `distribution.name < ROSDistribution.FOXY` should work. Really, we should add a `__lt__` or `__cmp__` method to `ROSDistribution`, and then we can just do `distribution < ROSDistribution.Foxy`. I'll add that to ROSWire now.

schmerl: Done.

ChrisTimperley (approved):
LGTM 👍🏻

PR #116, opened 2020-11-05
Update roswire to new ROS version API

schmerl:
There were recent changes made in `roswire` to handle both ROS1 and ROS2 distributions. These require some API changes to clients. This pull request reflects those changes.


ChrisTimperley (approved):
Looks good to me 👍🏻

PR #117, opened 2020-11-16
Add dynamic observation to rosdiscover

schmerl:
This pull request implements the scaffolding for `rosdiscover observe`, which takes a config and a container name or id and will dynamically get the architecture of the ROS system running inside the container.

The scaffolding in this PR does not have any functionality, but we can work out the high level design of the feature this way.

When this is done, let's not delete the branch but keep using it to develop the feature.


src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,48 @@
...
+from typing import Dict, Iterator
+
+from roswire import App, AppInstance, ROSVersion
+
+from .ros1 import ROS1ObserverConnection
+from .ros2 import ROS2ObserverConnection
+from ..config import Config
+from ..interpreter import NodeContext, SystemSummary
+
+_DEFAULT_URL = os.environ.get("DOCKER_HOST", "unix://var/run/docker.sock")
```
ChrisTimperley, line 14: Dead code?

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,48 @@
...
+        self._config = config
+        self.nodes: Dict[str, NodeContext] = {}
+
+    def summarise(self):
+        """Produces an immutable description of the system architecture."""
+        node_summaries = [node.summarise() for node in self.nodes.values()]
+        node_to_summary = {s.fullname: s for s in node_summaries}
+        return SystemSummary(node_to_summary)
+
+    def observe(self):
```
ChrisTimperley, line 41: Turn this into an abstract method. Implement in subclasses.

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,48 @@
...
+    @classmethod
+    @contextlib.contextmanager
+    def for_container(cls,
+                      container: str,
+                      config: Config,
+                      ) -> Iterator['Observer']:
+        """Constructs and interpreter for a given running container"""
+        app: App = App(config, None)
+        instance = app.attach(container, require_description=False)
+        yield Observer(instance, config)
```
ChrisTimperley, line 28: Generate appropriate `Observer` implementation based on type of ROS used by App.

src/rosdiscover/observer/observer.py
schmerl: Fixed

src/rosdiscover/observer/observer.py
schmerl: Fixed

src/rosdiscover/observer/observer.py
schmerl: Fixed

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+__all__ = ("Observer",)
```
ChrisTimperley, line 2: Minor nitpick: Can you add a blank line after `__all__`?

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,45 @@
...
+import contextlib
+from abc import ABC, abstractmethod
+from typing import Dict, Iterator
+import roswire
```
ChrisTimperley, line 6: Blank line between third-party imports and standard library imports

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,45 @@
...
+from roswire import AppInstance, ROSVersion
+
+from ..config import Config
```
ChrisTimperley, line 9: Since we don't actually need to interact with `Config` directly, we can do the old `if typing.TYPE_CHECKING` trick here. We will probably want to tweak `Config` to be a better OO class by adding `observer(...)` and `interpreter(...)` factory methods to it at some point in a future PR.

ChrisTimperley: Doing the above will also allow us to completely sidestep the nasty `ROS1Observer/ROS2Observer` import problem.

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,45 @@
...
+        app = rsw.app(config.image, config.sources)
+        instance = app.attach(container, require_description=True)
+        if app.description.distribution.ros == ROSVersion.ROS1:
+            from .ros1 import ROS1Observer
+            yield ROS1Observer(instance, config)
+        else:
+            from .ros2 import ROS2Observer
+            yield ROS2Observer(instance, config)
+
+    def __init__(self, app: AppInstance, config: Config):
```
ChrisTimperley, line 32: An `ABC` with a constructor should be avoided. ABCs should ideally be used to specify an interface rather than to provide an implementation.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+class ActionCandidate:
+    goal: Optional[Topic]
+    cancel: Optional[Topic]
+    status: Optional[Topic]
+    feedback: Optional[Topic]
+    result: Optional[Topic]
+
+    def __init__(self,
+                 action_name: str,
+                 is_server: bool):
```
ChrisTimperley, line 34: missing return type annotation

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+_RESULT = re.compile(r'(.*)/result')
+_RESULT_FMT = re.compile(r'(.*)Result')
+
+_NODES_TO_FILTER_OUT = {'/rosout'}
+_TOPICS_TO_FILTER_OUT = {'/rosout', '/rosout_agg'}
+_SERVICES_TO_FILTER_OUT = {'set_logger_level', 'get_loggers'}
+
+
+class ActionCandidate:
+    goal: Optional[Topic]
```
ChrisTimperley, line 26: `attrs` should be used to simplify this class.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+        self._action_name = action_name
+        self._server = is_server
+        self.goal = None
+        self.cancel = None
+        self.status = None
+        self.feedback = None
+        self.result = None
+
+    @property
+    def name(self):
```
ChrisTimperley, line 44: This seems redundant

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+        return self._action_name
+
+    @property
+    def fmt(self):
```
ChrisTimperley, line 48: missing type annotation

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+        assert self.is_complete()
+        return _GOAL_FMT.match(self.goal.format).group(1)
+
+    def is_complete(self):
```
ChrisTimperley, line 52: missing type annotation

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+        """Check if all information is there for an action.
+
+        Returns
+        ------_
```
ChrisTimperley, line 56: typo

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+        bool
+            Whether the action is complete
+        """
+        return self.goal is not None and self.cancel is not None and self.status is not None \
```
ChrisTimperley, line 60: This is a little difficult to parse. Consider splitting up.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+            and self.feedback is not None and self.result is not None
+
+    def add(self,
+            nc: NodeContext,
+            topic: Topic,
+            sub_if_server):
```
ChrisTimperley, line 66: missing type annotations

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+        Returns
+        ------_
+        bool
+            Whether the action is complete
+        """
+        return self.goal is not None and self.cancel is not None and self.status is not None \
+            and self.feedback is not None and self.result is not None
+
+    def add(self,
+            nc: NodeContext,
```
ChrisTimperley, line 64: prefer informative variable names

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+        else:
+            a_s = ActionCandidate(action_name, is_server)
+            existing[node][action_name] = a_s
+    else:
+        a_s = ActionCandidate(action_name, is_server)
+        existing[node] = {action_name: a_s}
+    return a_s
+
+
+def action_candidate(node: str,
```
ChrisTimperley, line 146: We rarely use module methods, so this feels a little out of place.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
+# -*- coding: utf-8 -*-
```
ChrisTimperley, line 1: There's a lot going on in this file. Consider breaking up the functionality into a number of smaller files.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+    format_match = _RESULT_FMT.match(fmt)
+    if result is not None and format_match is not None:
+        a_c = get_action_candidate(node, result.group(1), existing, publishes)
+        a_c.result = Topic(topic, fmt, False)
+        return True
+
+    return False
+
+
+def update_node_contexts_with_topics(nodecontexts, items, publishes,
```
ChrisTimperley, line 209: type annotations and docstrings

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+        Indicates whether the topic is published by the node (subscribed otherwise).
+    existing: Dict[str, Dict[str, ActionCandidate]]
+        The existing candidates. This will be updated if the topic might be part of
+        an action and the candidate isn't already known.
+    """
+    goal = _GOAL.match(topic)
+    format_match = _GOAL_FMT.match(fmt)
+
+    a_c = None
+    if goal is not None and format_match is not None:
```
ChrisTimperley, line 174: This method is pretty complex and the control flow is a little tricky to follow. It seems that there's a degree of duplicated code here. Is it possible to refactor this?

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+        a_s = ActionCandidate(action_name, is_server)
+        existing[node] = {action_name: a_s}
+    return a_s
+
+
+def action_candidate(node: str,
+                     topic: str,
+                     fmt: str,
+                     publishes: bool,
+                     existing: Dict[str, Dict[str, ActionCandidate]]):
```
ChrisTimperley, line 150: return type annotation

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+    """
+    Check if the topic might belong to an action. The parameter
+    :code:'existing' is updated with a new candidate if it is created. The topic is added to
```
ChrisTimperley, line 153: use `` to quote `:code:`

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,45 @@
...
+        self._app_instance = app
+        self._config = config
+        self._nodes: Dict[str, NodeContext] = {}
```
ChrisTimperley, line 35: This isn't really necessary or used.

src/rosdiscover/observer/ros2.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+from .observer import Observer
+
+
+class ROS2Observer(Observer):
+    def observe_and_summarise(self):
+        pass
```
ChrisTimperley, line 7: `raise NotImplementedError`

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def summarise(self):
+        """Produces an immutable description of the system architecture."""
+        node_summaries = [node.summarise() for node in self._nodes.values()]
+        node_to_summary = {s.fullname: s for s in node_summaries}
+        return SystemSummary(node_to_summary)
+
+    @abstractmethod
+    def observe_and_summarise(self):
```
ChrisTimperley, line 44: Rename to `observe`?

ChrisTimperley: Also add docstring and return type annotation

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            from .ros2 import ROS2Observer
+            yield ROS2Observer(instance, config)
+
+    def __init__(self, app: AppInstance, config: Config):
+        self._app_instance = app
+        self._config = config
+        self._nodes: Dict[str, NodeContext] = {}
+
+    def summarise(self):
```
ChrisTimperley, line 37: Return type annotation

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+                                f"goal({action_server.goal}), cancel({action_server.cancel}), "
+                                f"status({action_server.status}), feedback("
+                                f"{action_server.feedback}), result({action_server.result}")
+                            action_server.add_unfinished_to_context(nodecontexts[node])
+
+                for node, action_clients in action_client_candidates.items():
+                    for action_client in action_clients.values():
+                        if action_client.is_complete():
+                            nodecontexts[node].action_client(action_client.name, action_client.fmt)
+                            logger.warning(f"Action {action_client._action_name} is incomplete")
```
ChrisTimperley, line 293: This seems to be incorrect? Line 291 checks if `action_client` is complete.

src/rosdiscover/observer/ros1.py
ChrisTimperley: Better yet, the design here could be improved by splitting the different concerns of this class. One class can be a dumb, immutable data structure that represents some given `Action` and the other, `ActionBuilder`, can be used to gather the information necessary to build `Action` objects. If the builder lacks sufficient information, it can raise an exception.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,304 @@
...
+                                f"goal({action_client.goal}), cancel({action_client.cancel}), "
+                                f"status({action_client.status}), feedback("
+                                f"{action_client.feedback}), result({action_client.result}")
+
+                    else:
+                        action_client.add_unfinished_to_context(nodecontexts[node])
+            self._nodes = nodecontexts
+            return self.summarise()
+        except TimeoutError:
+            logger.exception("failed to connect to the ROS master. Is a robot running?")
```
ChrisTimperley, line 304: Probably missing a `raise` here. This will implicitly cause the method to return `None`.

src/rosdiscover/cli.py
```diff
@@ -64,6 +65,23 @@ def generate_acme(args):
             acme_gen.check_acme()
 
 
+def _observe(args):
```
ChrisTimperley, line 68: missing `-> None` return type annotation.

src/rosdiscover/cli.py
```diff
@@ -64,6 +65,23 @@ def generate_acme(args):
...
+    config = Config.from_yaml_string(args.config)
+    with Observer.for_container(args.container, config) as obs:
+        summary = obs.observe_and_summarise()
+    return summary
+
+
+def observe(args):
```
ChrisTimperley, line 75: missing `-> None` return type annotation.

src/rosdiscover/observer/observer.py
schmerl: I don't really agree with this, because this means would would need to either duplicate this code in each subclass or put it in some builder. Instead, I've made the `__init__` method abstract, as per: https://stackoverflow.com/questions/44800659/how-define-constructor-implementation-for-an-abstract-class-in-python

src/rosdiscover/observer/ros1.py
schmerl: Did this through attrs.

src/rosdiscover/observer/ros1.py
schmerl: done

src/rosdiscover/observer/ros1.py
schmerl: Done

src/rosdiscover/observer/ros1.py
schmerl: Done

src/rosdiscover/observer/ros1.py
schmerl: Done

src/rosdiscover/observer/ros1.py
schmerl: Done

src/rosdiscover/observer/ros1.py
schmerl: DOne

src/rosdiscover/observer/ros1.py
schmerl: Done

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+
+from loguru import logger
+from roswire import AppInstance
+
+from .observer import Observer
+from .ros1_actions import ActionCandidate, separate_topics_from_action
+from .. import Config
+from ..interpreter import NodeContext, ParameterServer, SystemSummary
+
+_NODES_TO_FILTER_OUT = {'/rosout'}
```
ChrisTimperley, line 12: Prefer (immutable) tuples over (mutable) sets for storing immutable sequences.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+                                               lambda node: nodecontexts[node].action_client,
+                                               nodecontexts)
+
+            self._nodes = nodecontexts
+            return self.summarise()
+        except TimeoutError:
+            logger.exception("failed to connect to the ROS master. Is a robot running?")
+            raise
+
+    def process_action_candidates(self,
```
ChrisTimperley, line 97: This seems like an implementation detail rather than part of the public API? Maybe it makes sense to change this to a protected method?

src/rosdiscover/cli.py
```diff
@@ -58,12 +59,29 @@ def generate_acme(args):
...
 
+def _observe(args) -> SystemSummary:
+    config = Config.from_yaml_string(args.config)
+    with Observer.for_container(args.container, config) as obs:
+        summary = obs.observe_and_summarise()
+    return summary
+
+
+def observe(args) -> None:
+    summary = _observe(args)
```
ChrisTimperley, line 76: I'm not sure why `_observe` is a separate method. It seems like the first three lines of `_observe` could be dropped inlined at this location.

src/rosdiscover/observer/observer.py
```diff
@@ -0,0 +1,51 @@
...
+from ..interpreter import NodeContext, SystemSummary
+
+if typing.TYPE_CHECKING:
+    from ..config import Config
+
+
+class Observer(ABC):
+
+    @classmethod
+    @contextlib.contextmanager
```
ChrisTimperley, line 21: There doesn't seem to be anything gained by using a context manager here? (I.e., there are no `try`/`finally` blocks, and so a context manager isn't necessary).

src/rosdiscover/observer/observer.py
ChrisTimperley: `self._nodes` isn't necessary here. We should try to ensure that `Observer` is a stateless class (since it needn't maintain state to implement its API). Maintaining internal state adds unnecessary complexity to the implementation and makes the code harder to test, debug, and maintain (method calls are no longer idempotent).

src/rosdiscover/observer/observer.py
ChrisTimperley: I'm not sure that this method belongs to the public API of `Observer`? The only API interaction that I can see is with an `.observe()` method that returns the instantaneous state of the associated system. Everything else is an implementation detail. (That's not to say that this method isn't useful, just that it might make more sense to change it to a protected method.)

Also, fitting with the above comment on statelessness, it might make sense to update this method to accept a `nodes` parameter rather than rely on `self._nodes`.

src/rosdiscover/observer/observer.py
ChrisTimperley: ^

src/rosdiscover/observer/observer.py
ChrisTimperley: Also, don't forget to add the return type annotation (`-> None`).

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+from ..interpreter import NodeContext, ParameterServer, SystemSummary
+
+_NODES_TO_FILTER_OUT = {'/rosout'}
+_TOPICS_TO_FILTER_OUT = {'/rosout', '/rosout_agg'}
+_SERVICES_TO_FILTER_OUT = {'set_logger_level', 'get_loggers'}
+
+
+class ROS1Observer(Observer):
+
+    def __init__(self, app: AppInstance, config: Config):
```
ChrisTimperley, line 19: There's no need to define this constructor.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+        ------
+        TimeoutException
+            If a connection cannot be made to the container
+        """
+        try:
+            nodecontexts: Dict[str, NodeContext] = {}
+            with self._app_instance.ros1() as ros:
+                nodes = [n for n in ros.nodes if n not in _NODES_TO_FILTER_OUT]
+                info = ros.state
+                # Create a context for each node
```
ChrisTimperley, line 41: To improve the readability of the code, it might help to introduce a separate method that creates the context for a particular node (e.g., `_fetch_node_context(node: Node) -> NodeContext)`.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+                # Places to store bits of actions, which only appear as topics
+                action_server_candidates: Dict[str, Dict[str, ActionCandidate]] = dict()
+                action_client_candidates: Dict[str, Dict[str, ActionCandidate]] = dict()
+                # Only include topics that we care about
+                significant_publishers = ROS1Observer._filter_out_topics_and_nodes(info.publishers)
+                significant_subscribers = ROS1Observer._filter_out_topics_and_nodes(
+                    info.subscribers)
+
+                # Process the topics
+                separate_topics_from_action(nodecontexts,
```
ChrisTimperley, line 66: Why are these methods part of the public module API?

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+                # Check if action candidates are complete (i.e., have all their topics)
+                # and add action if they are, or add the topics back in if they're not
+                self.process_action_candidates(action_server_candidates,
+                                               lambda node: nodecontexts[node].action_server,
+                                               nodecontexts)
+                self.process_action_candidates(action_client_candidates,
+                                               lambda node: nodecontexts[node].action_client,
+                                               nodecontexts)
+
+            self._nodes = nodecontexts
```
ChrisTimperley, line 91: Replace with a call to `summarise(node_contexts)`.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+            return self.summarise()
+        except TimeoutError:
```
ChrisTimperley, line 93: I'm not sure that this `try/except` is necessary since the exception _should_ contain an error message with roughly the same description. Removing the `try/except` would further help to improve readability.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+                    logger.debug(
+                        f"goal({action_candidate.goal}), cancel({action_candidate.cancel}), "
+                        f"status({action_candidate.status}), feedback("
+                        f"{action_candidate.feedback}), result({action_candidate.result}")
+                    action_candidate.add_unfinished_to_context(nodecontexts[node])
+
+    @staticmethod
+    def _filter_out_topics_and_nodes(
+            topics_and_nodes: Mapping[str, Collection[str]]
+    ) -> Dict[str, Collection[str]]:
```
ChrisTimperley, line 130: It would be helpful to add a very short docstring that describes the purpose of this method.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+        filtered: Dict[str, Collection[str]] = dict()
```
ChrisTimperley, line 131: prefer `{}` to `dict()`

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+                    logger.warning(f"Action {action_candidate.name} is incomplete")
+                    logger.debug(
+                        f"goal({action_candidate.goal}), cancel({action_candidate.cancel}), "
+                        f"status({action_candidate.status}), feedback("
+                        f"{action_candidate.feedback}), result({action_candidate.result}")
+                    action_candidate.add_unfinished_to_context(nodecontexts[node])
+
+    @staticmethod
+    def _filter_out_topics_and_nodes(
+            topics_and_nodes: Mapping[str, Collection[str]]
```
ChrisTimperley, line 129: This appears to be overindented?

ChrisTimperley: It may also be helpful to rename this parameter. `topics_and_nodes` sounds like a `Collection`, but seems to be a `Mapping` from something?

src/rosdiscover/observer/ros1_actions.py
```diff
@@ -0,0 +1,270 @@
+# -*- coding: utf-8 -*-
```
ChrisTimperley, line 1: It might be useful to add a brief module docstring to explain the contents and purpose of the module.

src/rosdiscover/observer/ros2.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+from roswire import AppInstance
+
+from .observer import Observer
+from .. import Config
+
+
+class ROS2Observer(Observer):
+
+    def __init__(self, app: AppInstance, config: Config):
```
ChrisTimperley, line 10: No need to define this constructor. This is already the implicit constructor.

src/rosdiscover/observer/ros2.py
```diff
@@ -0,0 +1,14 @@
...
+        super().__init__(app, config)
+
+    def observe_and_summarise(self):
```
ChrisTimperley, line 13: Missing return type annotation

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+        super().__init__(app, config)
+
+    def observe_and_summarise(self) -> SystemSummary:
+        """
+        Takes a snapshot of the ROS system running in the container referred to by
+        :code:``self.app_instance`` and returns the architecture summary for that system.
+
+        Returns
+        -------
+        The summary of the instantaneous state of the ROS system
```
ChrisTimperley, line 29: Missing return type:
```
SystemSummary
    The summary of ...
```

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class ROS1Observer(Observer):
+
+    def __init__(self, app: AppInstance, config: Config):
+        super().__init__(app, config)
+
+    def observe_and_summarise(self) -> SystemSummary:
+        """
+        Takes a snapshot of the ROS system running in the container referred to by
```
ChrisTimperley, line 24: Since `self.app_instance` is an implementation detail, we don't really need to expose it here. Instead, we can just say that this method observes the instantaneous state of the associated ROS system.

src/rosdiscover/observer/ros1.py
```diff
@@ -0,0 +1,135 @@
...
+        -------
+        The summary of the instantaneous state of the ROS system
+
+        Raises
+        ------
+        TimeoutException
+            If a connection cannot be made to the container
+        """
+        try:
+            nodecontexts: Dict[str, NodeContext] = {}
```
ChrisTimperley, line 37: `node_to_context` might better communicate that this is a dictionary rather than a list.

src/rosdiscover/observer/ros1_actions.py
```diff
@@ -0,0 +1,270 @@
...
+        The name of the topic that could be part of an action.
+    fmt: str
+        The topic format associated with the topic.
+    publishes: bool
+        Indicates whether the topic is published by the node (subscribed otherwise).
+    existing: Dict[str, Dict[str, ActionCandidate]]
+        The existing candidates. This will be updated if the topic might be part of
+        an action and the candidate isn't already known.
+    """
+    a_c = None
```
ChrisTimperley, line 169: Prefer informative names for readability (e.g., `action_client` or `client`).

src/rosdiscover/observer/ros1_actions.py
```diff
@@ -0,0 +1,270 @@
...
+            self.add(nc, self.feedback, False)
+        if self.result is not None:
+            self.add(nc, self.result, False)
+
+
+def get_action_candidate(node: str,
+                         action_name: str,
+                         existing: Dict[str, Dict[str, ActionCandidate]],
+                         is_server: bool) -> ActionCandidate:
+    """Gets the action candidate if it exsits
```
ChrisTimperley, line 114: Small typo: `exsits` -> `exists`.
This method also seems to create a new `ActionCandidate` if none exists rather than just trying to fetch one?

src/rosdiscover/observer/ros1_actions.py
```diff
@@ -0,0 +1,270 @@
...
+
+    Parameters
+    ----------
+    node: str
+        The name of the node that the action might belong to
+    action_name: str
+        The name of the action to look for or create
```
ChrisTimperley, line 121: To maintain consistent naming, it probably makes sense to either: (a) change `action_name` to `action`, or (b) change `node` to `node_name`.

src/rosdiscover/observer/ros1_actions.py
```diff
@@ -0,0 +1,270 @@
...
+    existing: Dict[str, Dict[str, ActionCandidate]]
+        The existing known actions. This is updated if a new candidate is created
+    is_server:
+        Whether this should be an action server
+
+    Returns
+    -------
+    ActionCandidate
+        The new or existing action candidate
+
```
ChrisTimperley, line 131: extra whitespace

src/rosdiscover/observer/ros1_actions.py
```diff
@@ -0,0 +1,270 @@
...
+    Parameters
+    ----------
+    node: str
+        The name of the node that the action might belong to
+    action_name: str
+        The name of the action to look for or create
+    existing: Dict[str, Dict[str, ActionCandidate]]
+        The existing known actions. This is updated if a new candidate is created
+    is_server:
+        Whether this should be an action server
```
ChrisTimperley, line 125: It's not clear what "this" is referring to.

src/rosdiscover/observer/ros1_actions.py
```diff
@@ -0,0 +1,270 @@
...
+_CANCEL = re.compile(r'(.*)/cancel')
+_STATUS = re.compile(r'(.*)/status')
+_FEEDBACK = re.compile(r'(.*)/feedback')
+_FEEDBACK_FMT = re.compile(r'(.*)Feedback')
+_RESULT = re.compile(r'(.*)/result')
+_RESULT_FMT = re.compile(r'(.*)Result')
+
+
+@attr.s(slots=True, auto_attribs=True)
+class ActionCandidate:
```
ChrisTimperley, line 25: I'm not 100% sure what it means for something to be an `ActionCandidate`. Can something be a candidate but not actually be a client or server?

src/rosdiscover/cli.py
schmerl: Am preparing for when we also want it to output Acme in addition to yaml. (There will be a smaller PR after this one to change the CLI to be more uniform about the Acme).

src/rosdiscover/observer/observer.py
schmerl: Perhaps I'm not understanding this properly, but line 79 of cli.py has this wrapped in a `with`

schmerl:
Withdrawing this pull request after a complete rewrite to trade readability over performance.

src/rosdiscover/observer/observer.py
schmerl: I was thinking that in the future we might want to be able to call this, which is why I made the class stateful.

src/rosdiscover/observer/ros2.py
schmerl: Actually, there seems to be a need to do this because the implicit constructor doesn't seem to get called (it gives a mypy error).

src/rosdiscover/observer/observer.py
schmerl: Done

src/rosdiscover/observer/observer.py
schmerl: Done.

src/rosdiscover/observer/observer.py
schmerl: Done

src/rosdiscover/observer/ros1.py
schmerl: Created a separate file for `NodeInfo`.

src/rosdiscover/observer/ros2.py
schmerl: Done

PR #130, opened 2021-07-21
Report missing models in component library

### LINKED ISSUE #128: Print a list of missing models prior to model composition
This should also be helpful for figuring out which models need to be recovered.

schmerl:
Resolves #128

### DIFF
--- a/src/rosdiscover/cli.py
+++ b/src/rosdiscover/cli.py
@@ -39,6 +39,10 @@ def launch(args) -> None:
     """Simulates the architectural effects of a `roslaunch` command."""
     summary = _launch_config(args)
     output = summary.to_dict()
+
+    # Warn of any undefined nodes
+    for node_summary in summary.unresolved:
+        print(f'Warning: {node_summary.name} in package: {node_summary.package} could not be found.')
     if args.output:
         with open(args.output, 'w') as f:
             yaml.dump(output, f, default_flow_style=False)
--- a/src/rosdiscover/interpreter/model.py
+++ b/src/rosdiscover/interpreter/model.py
@@ -28,7 +28,7 @@ class Model:
 
     @staticmethod
     def find(package: str, name: str) -> 'Model':
-        if (package, name) in Model._models.keys():
+        if (package, name) in Model._models:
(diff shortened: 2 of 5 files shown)


src/rosdiscover/cli.py
```diff
@@ -39,6 +39,10 @@ def launch(args) -> None:
     """Simulates the architectural effects of a `roslaunch` command."""
     summary = _launch_config(args)
     output = summary.to_dict()
+
+    # Warn of any undefined nodes
+    for node_summary in summary.unresolved:
```
ChrisTimperley, line 44: Doesn't this take place after the composition stage?

src/rosdiscover/interpreter/model.py
```diff
@@ -28,9 +28,9 @@ def register(package: str,
 
     @staticmethod
     def find(package: str, name: str) -> 'Model':
-        try:
+        if (package, name) in Model._models.keys():
```
ChrisTimperley, line 31: there shouldn't be a need for `.keys()` here. `(package, name) in Model._models` should suffice.

src/rosdiscover/interpreter/summary.py
```diff
@@ -90,3 +90,7 @@ def __getitem__(self, name: str) -> NodeSummary:
 
     def to_dict(self) -> List[Dict[str, Any]]:
         return [n.to_dict() for n in self.values()]
+
+    @property
+    def unresolved(self) -> Iterator[NodeSummary]:
```
ChrisTimperley, line 95: This does also seem useful, albeit orthogonal to #128

src/rosdiscover/cli.py
schmerl: I think so - but this it to do the final report of missing components (we were already logging them along the way.)

src/rosdiscover/interpreter/summary.py
schmerl: It is used in the reporting.

src/rosdiscover/interpreter/model.py
schmerl: Done.

PR #133, opened 2021-07-23
Extend config file to include information about where to find sources.

schmerl:
Config yaml files will now be able to specify where nodes's source files are in the image, e.g.,

```
image: therobotcooperative/fetch
sources:
- /opt/ros/melodic/setup.bash
- /ros_ws/devel/setup.bash
launches:
- /ros_ws/src/fetch_gazebo/fetch_gazebo/launch/pickplace_playground.launch
node_sources:
  - package: fetch
    node: fetch_gazebo
    sources:
    - fetch_gazebo.cc
 ```

### DIFF
--- a/src/rosdiscover/config.py
+++ b/src/rosdiscover/config.py
@@ -2,12 +2,65 @@
 __all__ = ('Config',)
 
 from types import MappingProxyType
-from typing import Any, Mapping, Sequence
+import typing as t
 
 from loguru import logger
 import attr
 import roswire
 import yaml
+
+@attr.s(frozen=True, slots=True, auto_attribs=True)
+class NodeSourceInfo:
+    """
+    This is a class to represent nodes and their sources, as a bootstrap for recovering node information
+    from static analysis.
+
+    Note: This is an interim class, and will be replaced by CMakelists.txt analysis in the fullness of time.
+
+    Attributes
+    ----------
+    package_name: str
+        The name of the package where the node is from
+    node_name: str
+        The name of the node provided by the package
+    sources: Sequence[str]
+        The list of sources for building the node
+    """
+    package_name: str
+    node_name: str
+    sources: t.Sequence[str]
+
+    @classmethod
+    def from_dict(cls, dict_: t.Mapping[str, t.Any]) -> 'NodeSourceInfo':
+        """
+        Raises
+        ------
+        ValueError
+            If 'package' is undefined.
+        ValueError
+            If 'node' is undefined.
+        ValueError
+            If 'sources' is undefined.
+        """
+        if 'package' not in dict_:
+            raise ValueError("'package' is undefined for the node source.")
+        if 'node' not in dict_:
+            raise ValueError("'node' is undefined for the node source.")
+        if 'sources' not in dict_:
+            raise ValueError("'sources' is undefined for the node source.")
+
+        if not isinstance(dict_['package'], str):
+            raise ValueError("expected 'package' to be a string")
+        if not isinstance(dict_['node'], str):
+            raise ValueError("expected 'node' to be a string")
+        if not isinstance(dict_['sources'], list):
+            raise ValueError("expected 'sources' to be a string")
+
+        return NodeSourceInfo(
+            package_name=dict_['package'],
+            node_name=dict_['node'],
+            sources=list(dict_['sources']),
+        )
 
 
 @attr.s(frozen=True, slots=True, auto_attribs=True)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/rosdiscover/config.py [hunk 1]
```diff
@@ -2,14 +2,68 @@
...
 import yaml
 
 
+@attr.s(frozen=True, slots=True, auto_attribs=True)
+class NodeSourceInfo:
+    """
+    This is a class to represent nodes and their sources, as a bootstrap for recovering node information
+    from static analysis.
+
+    Note: This is an interim class, and will be replaced by CMakelists.txt analysis in the fullness of time.
```
ChrisTimperley, line 19: Need this be an interim class? Could this not be data structure that's produced by the CMake analysis?

src/rosdiscover/config.py
```diff
@@ -2,14 +2,68 @@
...
+    package_name: str
+        The name of the package where the node is from
+    node_name: str
+        The name of the node provided by the package
+    sources: Sequence[str]
+        The list of sources for building the node
+    """
+    package_name: str
+    node_name: str
+    sources: t.Sequence[str]
```
ChrisTimperley, line 32: We'll also want to specify whether the node is a Python or C++ node (e.g., via an enum).

src/rosdiscover/config.py
```diff
@@ -2,14 +2,68 @@
...
+            raise ValueError("'node' is undefined for the node source.")
+        if 'sources' not in dict_:
+            raise ValueError("'sources' is undefined for the node source.")
+
+        if not isinstance(dict_['package'], str):
+            raise ValueError("expected 'package' to be a string")
+        if not isinstance(dict_['node'], str):
+            raise ValueError("expected 'node' to be a string")
+        if not isinstance(dict_['sources'], list):
+            raise ValueError("expected 'sources' to be a string")
```
ChrisTimperley, line 58: mismatch

src/rosdiscover/config.py
```diff
@@ -27,15 +81,18 @@ class Config:
         A set of environment variables that should be used by the application.
     app: roswire.app.App
         The ROSWire application for this configuration.
+    node_sources: List[NodeSourceInfo]
+        The list of information about the sources for nodes
```
ChrisTimperley, line 85: Maybe rewrite to: "Provides instructions on how to statically recover the architecture code for individual nodes."

src/rosdiscover/config.py
```diff
@@ -27,15 +81,18 @@ class Config:
...
     """
     image: str
-    sources: Sequence[str]
-    launches: Sequence[str]
-    environment: Mapping[str, str] = attr.ib(factory=dict)
+    sources: t.Sequence[str]
+    launches: t.Sequence[str]
+    environment: t.Mapping[str, str] = attr.ib(factory=dict)
+    node_sources: t.Mapping[t.Tuple[str, str], NodeSourceInfo] = attr.ib(factory=list)
```
ChrisTimperley, line 91: type mismatch: list is not a mapping

src/rosdiscover/config.py
schmerl: We haven't pulled in the ros-anatomy stuff to rosdiscover.

src/rosdiscover/config.py
```diff
@@ -62,14 +119,23 @@ def from_dict(cls, dict_: Mapping[str, Any]) -> 'Config':
...
+        sources: t.Sequence[str] = dict_['sources']
+        launches: t.Sequence[str] = dict_['launches']
+        environment: t.Mapping[str, str] = dict(dict_.get('environment', {}))
+        node_sources: t.Sequence[t.Dict[str, t.Any]] = list(dict_.get('node_sources', []))
+
         return Config(image=image,
                       sources=sources,
                       launches=launches,
-                      environment=environment)
+                      environment=environment,
```
ChrisTimperley, line 135: do you mind if we split the arguments to this function call across separate lines? readability becomes a little strained by line 136.

src/rosdiscover/config.py
schmerl: Let's defer this until we have the stuff from ros-anatomy, and do proper CMakelists.txt parsing. For now, because we're only integrating the C++ checking, let's just assume everything is C++

src/rosdiscover/config.py
schmerl: Done

src/rosdiscover/config.py
schmerl: DOne

src/rosdiscover/config.py
schmerl: Done

src/rosdiscover/config.py
schmerl: Done

ChrisTimperley (approved):

PR #141, opened 2021-07-26
Add launch file arguments to YML files

tobiasduerschmid:
This PR changes the YML scheme to:

```
launches:
- filename: /bug/src/repo/ros/src/util/packages/runtime_manager/scripts/launch_files/planning.launch
- filename: /bug/src/repo/ros/src/util/packages/runtime_manager/scripts/launch_files/map.launch
  arguments:
    tf_launch: /bug/src/repo/ros/src/util/packages/runtime_manager/scripts/launch_files/planning.launch
    other_arg: value
````

and passes the arguments to the launch file reader in roswire

### CODE REVIEW

File: src/rosdiscover/config.py [hunk 1]
```diff
@@ -57,14 +58,25 @@ def from_dict(cls, dict_: Mapping[str, Any]) -> 'Config':
             raise ValueError("expected 'sources' to be a list")
         if not isinstance(dict_['launches'], list):
             raise ValueError("expected 'launches' to be a list")
-
+        if isinstance(dict_['launches'][0], dict):
+            launch_args_provided = True
```
ChrisTimperley, line 62: This seems to suggest that if the first entry in launches is a YAML object (i.e., it provides args), then all entries that follow must also provide args? If so, that's probably being a little overly restrictive.

If the intention is to check that the items in `launches` are valid, you could do something like:
```python
if not all(type(d) in (dict, str) for d in dict_["launches"]):
# or if any(...) 
```

src/rosdiscover/config.py
```diff
@@ -57,14 +58,25 @@ def from_dict(cls, dict_: Mapping[str, Any]) -> 'Config':
...
+            raise ValueError("expected elements of 'launches' to be str or dict")
+            
         has_environment = 'environment' in dict_
         if has_environment and not isinstance(dict_['environment'], dict):
             raise ValueError("expected 'environment' to be a mapping")
 
         image: str = dict_['image']
         sources: Sequence[str] = dict_['sources']
-        launches: Sequence[str] = dict_['launches']
+        launches_inputs: Sequence[str] = dict_['launches']
```
ChrisTimperley, line 74: This type should probably be `Sequence[Any]` now (or a more specific albeit verbose union type).

src/rosdiscover/interpreter/interpreter.py
```diff
@@ -47,15 +48,15 @@ def summarise(self) -> SystemSummary:
         node_to_summary = {s.fullname: s for s in node_summaries}
         return SystemSummary(node_to_summary)
 
-    def launch(self, filename: str) -> None:
+    def launch(self, launchObj: Launch) -> None:
```
ChrisTimperley, line 51: stick to snakecase

src/rosdiscover/launch.py
```diff
@@ -0,0 +1,53 @@
...
+from types import MappingProxyType
+from typing import Any, Mapping, Sequence
+
+from loguru import logger
+import attr
+import yaml
+
+@attr.s(frozen=True, slots=True, auto_attribs=True)
+class Launch:
+    """
```
ChrisTimperley, line 13: One-line description of what this data structure does/describes would be helpful 👍🏻

src/rosdiscover/launch.py
```diff
@@ -0,0 +1,53 @@
...
+    filename: str
+        The file name of the launch files that should be launched
+    arguments: Mapping[str, str]
+        A set of launch arguments that should be passed to roslaunch
+    """
+    filename: str
+    arguments: Mapping[str, str]
+
+    def get_argv(self):
+        return list(map(lambda argk: f'{argk}:={self.arguments.get(argk)}', self.arguments.keys()))
```
ChrisTimperley, line 25: a list comprehension would be neater / easier to read here

src/rosdiscover/launch.py
```diff
@@ -0,0 +1,53 @@
...
+    ----------
+    filename: str
+        The file name of the launch files that should be launched
+    arguments: Mapping[str, str]
+        A set of launch arguments that should be passed to roslaunch
+    """
+    filename: str
+    arguments: Mapping[str, str]
+
+    def get_argv(self):
```
ChrisTimperley, line 24: don't forget to add a return type annotation

src/rosdiscover/launch.py
```diff
@@ -0,0 +1,53 @@
...
+
+        if not isinstance(dict_['filename'], str):
+            raise ValueError("expected 'filename' to be a string")
+
+        has_arguments = 'arguments' in dict_
+        if has_arguments and not isinstance(dict_['arguments'], dict):
+            raise ValueError("expected 'arguments' to be a mapping")
+
+        filename: str = dict_.get('filename')
+        logger.debug(f"FiLeNaMe: {type(filename)} {filename}")
```
ChrisTimperley, line 46: ?

src/rosdiscover/launch.py
```diff
@@ -0,0 +1,53 @@
...
+        arguments: Mapping[str, str] = dict(dict_.get('arguments', {}))
+        return Launch(filename=filename,
+                      arguments=arguments)
+
+    def __attrs_post_init__(self) -> None:
+        object.__setattr__(self, 'filename', self.filename)
```
ChrisTimperley, line 52: These two lines are dead code (e.g., this line is equivalent to `self.filename = self.filename`).  What's the intent here?

src/rosdiscover/config.py
```diff
@@ -8,6 +8,7 @@
 import attr
 import roswire
 import yaml
```
ChrisTimperley, line 10: bit of a nitpick that should be spotted by the linter: we split our import statements into three groups. First come the standard library imports, then third-party imports, and finally relative imports from our own package [https://www.python.org/dev/peps/pep-0008/#imports].

you should be able to run the linter locally while Travis misbehaves:
```
(someenv) $ tox
```

src/rosdiscover/config.py
tobiasduerschmid: fixed

src/rosdiscover/config.py
tobiasduerschmid: fixed

src/rosdiscover/interpreter/interpreter.py
tobiasduerschmid: fixed

src/rosdiscover/launch.py
tobiasduerschmid: done

src/rosdiscover/launch.py
tobiasduerschmid: done

src/rosdiscover/launch.py
tobiasduerschmid: done

src/rosdiscover/launch.py
tobiasduerschmid: some left-over debug-code. Removed it

src/rosdiscover/launch.py
tobiasduerschmid: Removed it.

ChrisTimperley:
It looks like this PR needs to be rebased on `master`.

src/rosdiscover/config.py
ChrisTimperley: tl;dr: add a new line between `import yaml` and `from .launch ...`

src/rosdiscover/launch.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+__all__ = ('Launch',)
+
+from typing import Any, Mapping, List
+
+import attr
+
+
+# Data class storing the file name of a launch file and its arguments
```
ChrisTimperley, line 9: move this description into the docstring on line 12

src/rosdiscover/interpreter/interpreter.py
```diff
@@ -47,15 +48,15 @@ def summarise(self) -> SystemSummary:
         node_to_summary = {s.fullname: s for s in node_summaries}
         return SystemSummary(node_to_summary)
 
-    def launch(self, filename: str) -> None:
+    def launch(self, launch_obj: Launch) -> None:
```
ChrisTimperley, line 51: maybe rename `launch_obj` to `launch_description`?

src/rosdiscover/launch.py
```diff
@@ -0,0 +1,47 @@
...
+            raise ValueError("'filename' is undefined in configuration")
+
+        if not isinstance(dict_['filename'], str):
+            raise ValueError("expected 'filename' to be a string")
+
+        has_arguments = 'arguments' in dict_
+        if has_arguments and not isinstance(dict_['arguments'], dict):
+            raise ValueError("expected 'arguments' to be a mapping")
+
+        filename: str = dict_.get('filename') or ""
```
ChrisTimperley, line 44: When/why would filename not exist? If there is such a case, prefer `None` over a sentinel value (i.e., the empty string) to ensure explicit corner case handling.

src/rosdiscover/config.py
tobiasduerschmid: done

src/rosdiscover/interpreter/interpreter.py
tobiasduerschmid: done

src/rosdiscover/launch.py
tobiasduerschmid: done

src/rosdiscover/launch.py
tobiasduerschmid: The filename should always exist. I only added this so that we can keep type checking bound to the type str. 
Otherwise, tox complains that it could be Optional[Any]

ChrisTimperley (approved):
LGTM 👍🏻

Issue #127, opened 2021-07-19
Add recover command for convenient static recovery of node models


ChrisTimperley:
```
$ rosdiscover recover some-config.yml package-name node-name
```

For now, we can assume that the user has the rosdiscover-cxx-extract Docker image (`rosdiscover/cxx-extract`) installed on their machine. From that container image, we will need to produce several Docker volumes that provide a portable/standalone installation of rosdiscover-cxx-extract.

The simplest approach to creating those volumes (`/opt/llvm11` and `/opt/rosdiscover`) is to first:

0. assign a fixed name to the volumes (e.g., `rosdiscover-cxx-extract-opt` and `rosdiscover-cxx-extract-llvm`).
1. check if the named volumes exist. If so, compare the volume's `ROSDISCOVER_IMAGE_SHA1` label against the SHA1 of the Docker image for rosdiscover-cxx-extract. If the SHA1 values are different (or if the named volumes don't exist), then we destroy the existing volumes (if any) and recreate new ones. (We also set the `ROSDISCOVER_IMAGE_SHA1` label during construction.)

For now, we can perform the above steps in the Makefile or install script for `rosdiscover-extract-cxx`. On the Python side, we should simply check that those volumes do indeed exist for now. Once those are in place, we continue as follows:

2. we provision the Docker image for the system-under-test (provided by the config file) and mount those named volumes as read-only volumes. (Right now, we can't do that directly in ROSWire. Either we can add that functionality, or we can simply create the container using Dockerblade and attach to it in ROSWire.)
3. we update the `PATH`, `LIBRARY_PATH`, and `LD_LIBRARY_PATH`, to include the `bin` and `lib` directories from the volumes.
4. we rebuild the catkin workspace and ensure that `compile_commands.json` is generated (by passing `-DCMAKE_EXPORT_COMPILE_COMMANDS=1` to the appropriate option in `catkin` or `catkin_make`). **(It would save a bunch of time and redundant computation if we didn't need to do this each time. We should think about ways to avoid this step or else minimise its cost.)**
5. we run `rosdiscover-cxx-extract` inside the container.
6. profit!

ChrisTimperley:
The first two steps in this approach have been implemented in rosdiscover-cxx-extract. You can build everything that you need to run the tool in any container via:
```
$ cd docker
$ make install
```

ChrisTimperley:
This has been implemented for a while.

PR #144, opened 2021-07-28
Implemented proof-of-concept static recovery command

ChrisTimperley:
This partially implements the functionality described in #127. Two things that are different and/or missing:

1. the static recovery program doesn't actually write the node summary to disk. This is simple to add, and I'll probably add an `--output` option to the `recover` command in rosdiscover, but for now, I just wanted to have a command that demonstrates that we can run the static recovery tool on arbitrary ROS systems.
2. ~~the actual command line interface is not what we will want to have eventually. Specifically, users need to provide the location of a Catkin workspace and the absolute paths of the (relevant) translation unit source files. We will want to move to simply "[package] [node]", but that relies on both CMake analysis and catkin version detection. (catkin_make and catkin tools produce `compile_commands.json` in different ways and store them to different places.)~~

**Update:** this command now accepts the following arguments:
```
$ rosdiscover recover [config] [package] [node] [sources, ...]
```

where sources are provided as paths relative to the package directory.

Example invocation:
```
$ rosdiscover recover example/chatter.yml roscpp_tutorials add_two_ints_client add_two_ints_client/add_two_ints_client.cpp
```


ChrisTimperley:
Example using an existing config file. Currently leads to a crash:
```
$ rosdiscover recover example/fetch.yml /ros_ws/build/fetch_depth_layer /ros_ws/src/fetch_ros/fetch_depth_layer/src/depth_layer.cpp
```

This is a bit of a weird case since this a dynamically loaded plugin rather than a node. Dealing with such plugins is currently out of scope, but at least you can run this command to see that the rest of the process is working.

ChrisTimperley:
A fun example that works using an image from The Robot Cooperative:
```
$ rosdiscover recover example/chatter.yml roscpp_tutorials add_two_ints_client add_two_ints_client/add_two_ints_client.cpp
```

(you'll need to rebuild the image with CMAKE_EXPORT_COMPILE_COMMANDS=1)

src/rosdiscover/cli.py
```diff
@@ -112,13 +121,27 @@ def main() -> None:
     parser = argparse.ArgumentParser(description=DESC)
     subparsers = parser.add_subparsers()
 
+    p = subparsers.add_parser(
+        'recover',
+        help='simulates the effects of a roslaunch.',
```
schmerl, line 126: This doesn't seem like the right help message for recover.

src/rosdiscover/recover/tool.py
```diff
@@ -1,20 +1,331 @@
...
+        build_directory = self._find_build_directory(workspace)
+        build_tool = self._detect_build_tool(workspace, build_directory)
+
+        compile_commands_directory: str
+        if build_tool == RosBuildTool.CATKIN_MAKE:
+            compile_commands_directory = build_directory
+        elif build_tool in (RosBuildTool.CATKIN_TOOLS, RosBuildTool.CATKIN_MAKE_ISOLATED):
+            compile_commands_directory = os.path.join(build_directory, package.name)
+        else:
+            assert (
```
schmerl, line 216: This assertion doesn't fail or print the message as far as I can see. Do you need `False` as the second argument?

src/rosdiscover/recover/tool.py
```diff
@@ -1,20 +1,331 @@
...
 
-    def recover(self, node_type: str, package: str) -> RecoveredNodeModel:
-        raise NotImplementedError
+        env = {
+            "PATH": "/opt/rosdiscover/bin:${PATH:-}",
+            "LIBRARY_PATH": "/opt/rosdiscover/lib:${LIBRARY_PATH:-}",
+            "LD_LIBRARY_PATH": "/opt/rosdiscover/lib:${LD_LIBRARY_PATH:-}",
+        }
+        env_args = [f"{var}={val}" for (var, val) in env.items()]
+        args = env_args + [
```
schmerl, line 320: Not sure what `rosdiscover -p` does. Maybe a comment about this would help. (e.g., I am thinking that it is calling the cli of this code, but it doesn't seem to be doing that.)

src/rosdiscover/recover/tool.py
ChrisTimperley: I'm mixing up Python and C++ :-) We can either do `assert False, msg` or `raise AssertionError(msg)`

schmerl: I think raise AssertionError, or ValueError.

src/rosdiscover/recover/tool.py
ChrisTimperley: The binary name is a little misleading. I've renamed this to `rosdiscover-cxx-extract`. This line is essentially calling the static recovery binary inside the container. `-p` specifies the directory that holds the `compile_commands.json` file.

schmerl (approved):

PR #145, opened 2021-08-02
Add code to check if information about source code is in the config

schmerl:


src/rosdiscover/config.py
```diff
@@ -90,7 +90,7 @@ class Config:
     sources: t.Sequence[str]
     launches: t.Sequence[Launch]
     environment: t.Mapping[str, str] = attr.ib(factory=dict)
-    node_sources: t.Mapping[t.Tuple[str, str], NodeSourceInfo] = attr.ib(factory=dict)
+    node_sources: t.Sequence[NodeSourceInfo] = attr.ib(factory=list)
```
ChrisTimperley, line 93: suggest `t.Collection` if order isn't important

PR #146, opened 2021-08-02
Added infrastructure for representing and caching recovered node models

ChrisTimperley:


src/rosdiscover/recover/model.py
```diff
@@ -0,0 +1,55 @@
...
+import typing as t
+
+import attr
+
+from ..interpreter import NodeContext
+
+
+@attr.s(frozen=True, slots=True)
+class RecoveredNodeModel:
+    """Provides a description of the run-time architecture of a given node.
```
schmerl, line 14: It looks like this class just stores information about the pre-analyzed node, rather than information about the pubs, subs, etc. that is the run-time architecture. So, I think the description is right, but the fields are wrong?

schmerl (changes requested):

src/rosdiscover/recover/model.py
ChrisTimperley: The fields should be okay. The description is a little misleading, though. The description makes it sound like we're providing the concrete node model for a particular configuration, whereas what we're actually doing is providing a symbolic program that can be used to obtain a concrete set of pubs, subs, etc. using the interpreter.

ChrisTimperley: Improved description to eliminate ambiguity

PR #152, opened 2021-08-09
Added interface for fetching models for a given project

ChrisTimperley:
Note that, for now, model recovery is explicitly disabled. Once we can load recovered models from their JSON description, we can drop the `allow_recovery=False` option.

### DIFF
--- a/src/rosdiscover/project/models.py
+++ b/src/rosdiscover/project/models.py
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
 __all__ = ("ProjectModels",)
 
+import typing as t
+
 import attr
 
 from ..config import Config
 from ..interpreter import NodeModel, HandwrittenModel
-from ..recover.database import RecoveredModelDatabase
+from ..recover import NodeRecoveryTool, RecoveredModelDatabase
 
 
 @attr.s(slots=True, auto_attribs=True)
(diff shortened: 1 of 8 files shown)


src/rosdiscover/project/models.py
```diff
@@ -0,0 +1,135 @@
...
+        package: str
+            The name of the package to which the node belongs.
+        node: str
+            The name of the type of node.
+
+        Returns
+        -------
+        NodeModel
+            When searching for a given model, highest preference will be given to
+            handwritten models. If there is no recovered model, then this method will
```
schmerl, line 108: I think you mean priority will be given to recovered models?

schmerl (approved):
LGTM. Just one comment to correct the comments section

src/rosdiscover/project/models.py
```diff
@@ -0,0 +1,135 @@
...
+            If there is no model could be fetched for the given node and placeholders
+            have been disabled.
+        """
+        # TODO this exists to make it easier to customize preferences later on
+        # e.g., to prefer recovered models over handwritten ones
+        model_sources: t.List[t.Callable[[str, str], t.Optional[NodeModel]]] = [
+            self._fetch_handwritten,
+        ]
+        if self.allow_recovery:
+            model_sources.append(self._recover)
```
schmerl, line 124: Shouldn't this be prepended the list to give it priority?

schmerl (approved):
LGTM. One more question,

src/rosdiscover/project/models.py
ChrisTimperley: For now, I've given priority to handwritten models since they're likely to be more accurate in the cases where we have them. I was thinking about adding an option to this class to switch preferences between recovered and handwritten (and maybe `allow_handwritten`, too), but I'll probably leave that to a future PR.

PR #153, opened 2021-08-11
Added loading and storage of recovered node models into the rest of ROSDiscover

ChrisTimperley:
Note that this PR does not provide an interpreter for the recovered models. I'll spin that up on a separate PR to keep things moving faster.

To quickly demonstrate this functionality:
```
$ pipenv run rosdiscover launch example/chatter.yml
```

### DIFF
--- a/example/chatter.yml
+++ b/example/chatter.yml
@@ -3,4 +3,13 @@ sources:
   - /opt/ros/kinetic/setup.bash
   - /ros_ws/devel/setup.bash
 launches:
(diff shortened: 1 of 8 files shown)


schmerl (approved):
Actions are missing. Is the plan to add them in a separate pull request?

ChrisTimperley:
> 
> 
> Actions are missing. Is the plan to add them in a separate pull request?

Good catch. Actions aren't part of the NodeHandle API, so they haven't been added yet. I'll create an issue to keep track of that feature.

PR #154, opened 2021-08-11
Provide more direct access to the ROS distribution in the context

schmerl:
This could previously be gotten via the `app` property, but to make it easier to use in component models, we provide a shortcut.

### DIFF
--- a/src/rosdiscover/models/joint_state_publisher.py
+++ b/src/rosdiscover/models/joint_state_publisher.py
@@ -1,8 +1,10 @@
-from ..interpreter import model
+from roswire import ROSDistribution
+
+from ..interpreter import model, NodeContext
 
 
 @model('joint_state_publisher', 'joint_state_publisher')
-def joint_state_publisher(c):
+def joint_state_publisher(c: NodeContext):
     # https://github.com/ros/joint_state_publisher/blob/kinetic-devel/joint_state_publisher/joint_state_publisher/joint_state_publisher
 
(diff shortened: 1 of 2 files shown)


src/rosdiscover/models/joint_state_publisher.py
```diff
@@ -19,7 +21,9 @@ def get_param(name, value=None):
     get_param("publish_default_positions", True)
     get_param("publish_default_velocities", False)
     get_param("publish_default_efforts", False)
-    get_param("use_gui", False)
+
+    if c.ros_distro < ROSDistribution.NOETIC:
```
ChrisTimperley, line 25: how does ordering work for ROS1 vs. ROS2?

ChrisTimperley (approved):
LGTM. I made a small comment about ROS1 vs. ROS2 comparisons that we might want to think about.

src/rosdiscover/models/joint_state_publisher.py
schmerl: This is the definition for `lt` for `ROSDistribution`:

```
    def __lt__(self, other: Any) -> int:
        if not isinstance(other, ROSDistribution):
            m = f"can only compare ROSDistribution objects"
            raise ValueError(m)
        if self.ros != other.ros:
            m = f"can only compare ROSDistributions for same version"
            raise ValueError(m)
        return self.display_name < other.display_name
```

So I think we're ok.

PR #159, opened 2021-08-20
Added config node source nodelet indication

schmerl:
Config file now specifies if the node is a "normal" node or a "nodelet".

### DIFF
--- a/src/rosdiscover/config.py
+++ b/src/rosdiscover/config.py
@@ -14,7 +14,7 @@ from .launch import Launch
 
 
 class ROSNodeKind(enum.Enum):
-    NORMAL = "normal"
+    NODE = "node"
     NODELET = "nodelet"
 
     @classmethod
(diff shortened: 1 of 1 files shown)


schmerl:
Note that we could make `kind` optional and default to `NORMAL`.

src/rosdiscover/config.py
```diff
@@ -12,6 +13,19 @@
 from .launch import Launch
 
 
+class ROSNodeKind(enum.Enum):
+    NORMAL = "normal"
```
ChrisTimperley, line 17: maybe just `node` rather than `normal`?

ChrisTimperley (approved):
Looks good to me. I made a minor comment about node vs. normal.

schmerl:
> 
> 
> Note that we could make `kind` optional and default to `NORMAL`.

Made this change - default is now `NODE`.

PR #160, opened 2021-08-20
Move body of Acme processing into separate module to control export in __init__.py

schmerl:
Refactored the Acme code into a separate file, rather than have everything in init.

Also update the acme-lib-ros.jar to improve error reporting.


ChrisTimperley (approved):
Looks fine to me. If you want to make unilateral changes to the Acme code, go ahead 👍🏻

PR #162, opened 2021-08-23
Add entrypoint specification for node recovery

schmerl:
The entrypoint for a node could be `main` or `onInit`. So, specify this in the config file.

### DIFF
--- a/src/rosdiscover/config.py
+++ b/src/rosdiscover/config.py
@@ -96,7 +96,7 @@ class NodeSourceInfo:
             package_name=dict_['package'],
             node_name=dict_['node'],
             node_kind=kind,
-            entrypoint=dict_['entrypoint'],
+            entrypoint=dict_.get('entrypoint'),
             sources=list(dict_['sources'])
         )
 
--- a/src/rosdiscover/config.py
+++ b/src/rosdiscover/config.py
@@ -42,7 +42,7 @@ class NodeSourceInfo:
         The name of the node provided by the package
     node_kind: ROSNodeKind
         The kind of node represented
-    entry_point: str
+    entrypoint: str
         The entry point for the main program in the source. The format for this is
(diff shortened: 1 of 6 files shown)


src/rosdiscover/config.py
```diff
@@ -86,6 +95,7 @@ def from_dict(cls, dict_: t.Mapping[str, t.Any]) -> 'NodeSourceInfo':
             package_name=dict_['package'],
             node_name=dict_['node'],
             node_kind=kind,
+            entry_point=dict_['entry_point'],
```
ChrisTimperley, line 98: should we do `dict_.get("entry_point")` to avoid the need to specify entrypoints for normal nodes?

src/rosdiscover/recover/symbolic.py
```diff
@@ -349,34 +352,36 @@ class SymbolicProgram:
...
         return {
             "program": {
+                "entry_point": self.entry_point,
                 "functions": [f.to_dict() for f in self.functions.values()],
             },
         }
 
     @property
-    def main(self) -> SymbolicFunction:
+    def entry_fn(self) -> SymbolicFunction:
```
ChrisTimperley, line 381: Minor point: Can we change this to `entrypoint`?

src/rosdiscover/recover/symbolic.py
```diff
@@ -341,6 +341,9 @@ class SymbolicProgram:
 
     Attributes
     ----------
+    entry_point: str
+        The name of the function that serves as the entry point for the
+        proagram.
```
ChrisTimperley, line 346: typo

ChrisTimperley (approved):
This looks great! Thanks! 👍🏻 If it isn't too much trouble, can we rename `entry_point` to `entrypoint`? No worries if it's a nuisance to change.

PR #176, opened 2021-09-09
Added models for AutoRally Python nodes

schmerl:


ChrisTimperley (approved):
These are really hard to manually extract 😅 But everything looks good. Thanks for taking the time to put this together! 👍🏻

PR #179, opened 2021-09-13
CMake parsing to process source nodes.

schmerl:


src/rosdiscover/project/models.py
```diff
@@ -60,30 +61,37 @@ def close(self) -> None:
 
     def _recover(self, package: str, node: str) -> t.Optional[NodeModel]:
         # have we already recovered this model?
+        print(f"Recovering {package}/{node}")
```
ChrisTimperley, line 64: prefer logger calls

src/rosdiscover/project/models.py
```diff
@@ -60,30 +61,37 @@ def close(self) -> None:
...
-            package,
-            node,
-            node_info.entrypoint,
-            node_info.sources,
-            node_info.restrict_to_paths
-        )
+            try:
+                logger.info(f"Attempting to recover {package}/{node} from CMakeLists.txt")
+                model = self._recovery_tool.recover_using_cmakelists(package, node)
+            except ValueError as e:
```
ChrisTimperley, line 73: We may also want to catch `RuntimeError` here, too. (We use that in cases where the input is fine but the C++ static recovery failed.)

src/rosdiscover/project/models.py
```diff
@@ -60,30 +61,37 @@ def close(self) -> None:
...
+                logger.error(f"Error recovering {package}/{node}: {e}")
```
ChrisTimperley, line 74: We can use `logger.exception` to record the stack trace here. (We can then drop `{e}` from the message.)

src/rosdiscover/project/models.py
schmerl: Done

src/rosdiscover/project/models.py
schmerl: Done

src/rosdiscover/project/models.py
schmerl: Done.

PR #181, opened 2021-09-14
Updated CLI recover command to use CMakeLists.txt parsing

schmerl:


ChrisTimperley (changes requested):
It looks like we need to update our Pipfile.lock to reflect recent changes to the Pipfile. (Without those changes, we end up using an older, incompatible version of ROSWire, which leads to a crash.)

Otherwise this looks great! 👍🏻

PR #194, opened 2021-09-22
Change `observe` to take a duration and not a number of iterations

schmerl:


src/rosdiscover/cli.py
```diff
@@ -115,32 +116,38 @@ def _observe(args) -> SystemSummary:
     return summary
 
 
-def _periodic_observe(period: int, args: argparse.Namespace) -> SystemSummary:
+def _periodic_observe(interval: int, args: argparse.Namespace) -> SystemSummary:
```
ChrisTimperley, line 119: should interval be a float?

src/rosdiscover/cli.py
```diff
@@ -115,32 +116,38 @@ def _observe(args) -> SystemSummary:
...
+                go = stopwatch.duration < args.duration
+            logger.debug('Sleeping')
+            time.sleep(interval)
+            if 'duration' in args:
+                go = stopwatch.duration < args.duration
         except KeyboardInterrupt:
             go = False
-    logger.info(f"Finished observing - {iterations+1} observations in total.")
+    stopwatch.stop()
+    logger.info(f"Finished observing - {iterations+1} observations in total after {stopwatch.duration} seconds.")
```
ChrisTimperley, line 145: It would be nice to use decimal formatting (e.g., `{stopwatch.duration:.3f}`) to display the duration

src/rosdiscover/cli.py
ChrisTimperley: bit of a nitpick, interval could just be grabbed from `args`.

ChrisTimperley (approved):
Other than a couple of minor points, this looks good to me! 👍🏻

src/rosdiscover/cli.py
schmerl: We could. I wanted to make it clear through the signature a bit what was required for it to work.

src/rosdiscover/cli.py
schmerl: Done.

PR #195, opened 2021-09-22
Allow `observe` to run a script on container

schmerl:


src/rosdiscover/observer/observer.py
```diff
@@ -59,3 +60,20 @@ def observe(self) -> SystemSummary:
...
+
+        Parameters
+        ----------
+        path_on_host: str
+            The path to the script on the host (commands in the script should
+            be relative to the container, not the host).
+
+        Returns
+        -------
+        int
```
ChrisTimperley, line 76: doesn't match return type

ChrisTimperley: also, this method probably doesn't need to be abstract. I think that you could getaway with the same method for ROS1 and ROS2?

src/rosdiscover/observer/ros1.py
```diff
@@ -94,3 +97,17 @@ def _transform_state_to_nodeinfo(self, state: SystemState) -> Collection[NodeInf
                         reorganized_nodes[node_name].provides.add(service)
 
         return list(reorganized_nodes.values())
+
+    def execute_script(self, path_on_host: str) -> Popen:
+        if not os.path.exists(path_on_host):
+            raise FileNotFoundError(f"'{path_on_host}' not found.")
+        assert self._app_instance is not None
+
+        path_on_container = os.path.join('/tmp', os.path.basename(path_on_host))
```
ChrisTimperley, line 106: We could use this to be safer: https://github.com/ChrisTimperley/dockerblade/blob/master/src/dockerblade/files.py#L540

src/rosdiscover/observer/ros1.py
```diff
@@ -94,3 +97,17 @@ def _transform_state_to_nodeinfo(self, state: SystemState) -> Collection[NodeInf
...
+        self._app_instance.files.copy_from_host(path_on_host, "/launch_extractor.py")
```
ChrisTimperley, line 107: I think that the second argument should be `path_on_container`

src/rosdiscover/observer/ros1.py
```diff
@@ -94,3 +97,17 @@ def _transform_state_to_nodeinfo(self, state: SystemState) -> Collection[NodeInf
...
+
+        cmd = path_on_container
```
ChrisTimperley, line 109: We could do `bash {path_on_container}` to avoid assuming the correct shebang and permissions?

src/rosdiscover/observer/observer.py
schmerl: Agreed. Done.

src/rosdiscover/observer/ros1.py
schmerl: Done.

src/rosdiscover/observer/ros1.py
schmerl: Yes. Done.

src/rosdiscover/observer/ros1.py
schmerl: Done.

src/rosdiscover/observer/ros2.py
```diff
@@ -1,9 +1,10 @@
...
+
 from .observer import Observer
 from ..interpreter import SystemSummary
 
 
 class ROS2Observer(Observer):
 
     def observe(self) -> SystemSummary:
-        raise NotImplementedError
+        raise NotImplementedError()
```
ChrisTimperley, line 10: no need for parens

src/rosdiscover/observer/ros2.py
```diff
@@ -1,9 +1,10 @@
 # -*- coding: utf-8 -*-
+
```
ChrisTimperley, line 2: any reason for the new line?

ChrisTimperley (approved):
LGTM sans a couple of very pedantic issues 👍🏻

PR #197, opened 2021-09-23
Add options to launch launch scripts from the config file when doing `observe`

schmerl:


schmerl:
BTW, `Popen.terminate` and `Popen.kill` in dockerblade don't seem to be working. At least, the don't seem to be working the way the command line does. 

I get the tracing:
```
2021-09-23 17:17:54.246 | INFO     | rosdiscover.cli:_periodic_observe:180 - Killing bash /tmp/tmp.RgfVxCnwQL.sh
2021-09-23 17:17:54.249 | INFO     | rosdiscover.cli:_periodic_observe:180 - Killing bash /tmp/tmp.PokDaCA8aV.sh
2021-09-23 17:17:54.267 | INFO     | rosdiscover.cli:_periodic_observe:180 - Killing bash /tmp/tmp.SEAp09MraL.sh
```
But when I open a shell, and do `ps ux` I see:
```
root      8821  0.0  0.0  20792  4180 pts/6    Ss+  17:15   0:00 bash /tmp/tmp.RgfVxCnwQL.sh
root     10056  0.0  0.0  20792  4152 pts/7    Ss+  17:16   0:00 bash /tmp/tmp.PokDaCA8aV.sh
root     10056  0.0  0.0  20792  4152 pts/7    Ss+  17:16   0:00 bash /tmp/tmp.PokDaCA8aV.sh
```
(As well as all the ros nodes). If from a terminal on the container I do `kill 8821` etc, that stops the processes and the child processes.

src/rosdiscover/cli.py
```diff
@@ -166,10 +171,14 @@ def _periodic_observe(interval: float, args: argparse.Namespace) -> SystemSummar
         if process and not process.poll():
             logger.warning(f"Observations seemed to have stopped before '{args.run_script}' completed.")
 
-        logger.info(f"Finished observing - {iterations+1} observations in total after {stopwatch.duration} seconds.")
+        logger.info(f"Finished observing - {iterations+1} observations in total after {stopwatch.duration:.3g} "
```
ChrisTimperley, line 174: why `g` (which is sometimes a lowercase exponential format) rather than `f`?

schmerl: typo

schmerl: Which funnily enough produces the same output in this case.

src/rosdiscover/observer/ros1.py
```diff
@@ -94,3 +100,44 @@ def _transform_state_to_nodeinfo(self, state: SystemState) -> Collection[NodeInf
...
+        app_instance = self._app_instance
+        for launch in self._config.launches:
+            if not app_instance.files.exists(launch.filename):
+                raise FileNotFoundError(launch.filename)
+
+        processes = []
+
+        for launch in self._config.launches:
+            launch_script = self._generate_launch_script_on_host(launch)
+            cmd = f"bash {launch_script}"
```
ChrisTimperley, line 115: prefer /bin/bash to be a bit more robust against bad PATH variables

src/rosdiscover/cli.py
schmerl: Done

src/rosdiscover/observer/ros1.py
schmerl: Done

ChrisTimperley (approved):
LGTM 👍🏻

PR #213, opened 2021-10-20
Log Acme errors in addition to printing them.

schmerl:


schmerl:
Merging myself because this is in the Acme part.

PR #232, opened 2021-11-01
Fixed various turtlebot issues

schmerl:


src/rosdiscover/interpreter/interpreter.py
```diff
@@ -273,21 +274,31 @@ def _load(self,
 
         try:
             model = self.models.fetch(pkg, nodetype)
+            if isinstance(model, PlaceholderModel) and name != nodetype:
+                model = self.models.fetch(pkg, name)
```
ChrisTimperley, line 278: might be worth adding a note about why `name` is used here

src/rosdiscover/models/amcl.py
```diff
@@ -1,10 +1,10 @@
 # -*- coding: utf-8 -*-
-from ..interpreter import model
+# from ..interpreter import model
 
 M_PI = 3.14159265358979323846
 
 
-@model('amcl', 'amcl')
+# @model('amcl', 'amcl')
```
ChrisTimperley, line 7: ?

src/rosdiscover/models/map_server.py
```diff
@@ -1,7 +1,7 @@
-from ..interpreter import model
+# from ..interpreter import model
 
 
-@model('map_server', 'map_server')
+# @model('map_server', 'map_server')
```
ChrisTimperley, line 4: ?

src/rosdiscover/interpreter/interpreter.py
schmerl: Done

src/rosdiscover/models/amcl.py
schmerl: This is commented out so that we attempt to do recovery. Didn't want to delete in case we need it again.

schmerl: Done, with comment indicating why.

src/rosdiscover/models/map_server.py
schmerl: Done with comments justifying why.

ChrisTimperley (approved):

PR #240, opened 2021-11-06
Error reproducer

schmerl:
Add a error_reproducer handwritten model for reproducing some of the errors for the misconfiguration database in the cases where we can't detect them from the container because there are no users of the misconfigured things.

### DIFF
--- a/src/rosdiscover/models/rosdiscover_error_reproducer.py
+++ b/src/rosdiscover/models/rosdiscover_error_reproducer.py
@@ -1,0 +1,11 @@
+# -*- coding: utf-8 -*-
+from ..interpreter import model, NodeContext
+
+@model('*', 'error_reproducer')
+def error_reproducer(c: NodeContext):
+    topics = c.read("~topic", [])
+    for t in topics:
+        if t["direction"] == 'pub':
+            c.pub(t["name"], t["type"])
+        elif t["direction"] == 'sub':
+            c.sub(t["name"], t["type"])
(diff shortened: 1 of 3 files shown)


src/rosdiscover/models/rosdiscover_error_reproducer.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+from ..interpreter import model, NodeContext
+
+
+@model('*', 'error_reproducer')
```
ChrisTimperley, line 5: It might be worth adding a docstring to this method or module that describes the intent.

src/rosdiscover/project/models.py
```diff
@@ -97,6 +97,8 @@ def _fetch_handwritten(self, package: str, node: str) -> t.Optional[NodeModel]:
         logger.debug(f"Handwritten {package}/{node}")
         if HandwrittenModel.exists(package, node):
             return HandwrittenModel.fetch(package, node)
+        if HandwrittenModel.exists("*", node):
```
ChrisTimperley, line 100: what does this mean?

schmerl: Added comment.

src/rosdiscover/models/rosdiscover_error_reproducer.py
schmerl: Done.

ChrisTimperley (approved):
👍🏻 LGTM

PR #242, opened 2021-11-08
Normalize nodelet manager name

schmerl:


src/rosdiscover/interpreter/interpreter.py
```diff
@@ -290,6 +290,9 @@ def _load(self,
         if args.startswith('manager'):
             # This is being loaded into an existing manager, so find that as the context
             manager_name = args.split(" ")[1]
+            if namespace:
+                manager_name = f"{namespace}/{manager_name}"
```
ChrisTimperley, line 294: let's talk about this. I have a case where it will break.

PR #243, opened 2021-11-08
Added handling of file argument to spawn_model.py

ChrisTimperley:


src/rosdiscover/models/spawn_model.py
```diff
@@ -8,11 +11,50 @@
...
+        logger.error(f"unable to load URDF XML from file [{filename}]: expected absolute path")
+        return None
+
+    logger.debug(f"loading URDF XML from file [{filename}]")
+    urdf_contents = c.read_file(filename)
+    logger.debug(f"loaded URDF XML contents from file [{filename}]:\n{urdf_contents}")
+    logger.debug("parsing URDF XML contents")
+    xml = ET.fromstring(urdf_contents)
+    logger.debug("parsed UDF XML contents")
+    return xml
```
schmerl, line 49: This might have the same problem as the other read, i.e., the urdf isn't surrounded by <robot>,

schmerl (approved):

src/rosdiscover/models/spawn_model.py
ChrisTimperley: From looking at one particular file, it seems to be well-formed, but yeah, I suppose that the same issue could occur. I'll tweak the code.

PR #253, opened 2022-06-29
Add controlflow

tobiasduerschmid:
This PR adds if and while to the python-side of RosDiscover

### DIFF
--- a/src/rosdiscover/recover/loader.py
+++ b/src/rosdiscover/recover/loader.py
@@ -229,11 +229,11 @@ class SymbolicProgramLoader:
         else:
             raise ValueError(f"unknown statement kind: {kind}")
 
-    def _load_while(self, dict_: t.Mapping[str, t.Any]) -> SymbolicCompound:
+    def _load_while(self, dict_: t.Mapping[str, t.Any]) -> SymbolicWhile:
         assert dict_["kind"] == "while"
         return SymbolicWhile(self._load_compound(dict_["body"]), self._load_value(dict_["condition"]))
 
-    def _load_if(self, dict_: t.Mapping[str, t.Any]) -> SymbolicCompound:
+    def _load_if(self, dict_: t.Mapping[str, t.Any]) -> SymbolicIf:
         assert dict_["kind"] == "if"
         return SymbolicIf(
             true_body=self._load_compound(dict_["trueBranchBody"]),
--- a/src/rosdiscover/recover/loader.py
+++ b/src/rosdiscover/recover/loader.py
@@ -63,6 +63,10 @@ class SymbolicProgramLoader:
             name=name,
             type_=type_,
         )
+    
+    def _load_bool_literal(self, dict_: t.Mapping[str, t.Any]) -> BoolLiteral:
+        assert dict_["kind"] == "bool-literal"
+        return BoolLiteral(dict_["literal"])
 
     def _load_string_literal(self, dict_: t.Mapping[str, t.Any]) -> StringLiteral:
         assert dict_["kind"] == "string-literal"
@@ -98,6 +102,8 @@ class SymbolicProgramLoader:
             return self._load_arg(dict_)
         elif kind == "string-literal":
             return self._load_string_literal(dict_)
+        elif kind == "bool-literal":
+            return self._load_bool_literal(dict_)            
         elif kind == "node-handle":
             return self._load_node_handle(dict_)
         elif kind == "variable-reference":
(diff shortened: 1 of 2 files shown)


src/rosdiscover/recover/loader.py
```diff
@@ -220,9 +222,24 @@ def _load_statement(self, dict_: t.Mapping[str, t.Any]) -> SymbolicStatement:
...
+    def _load_while(self, dict_: t.Mapping[str, t.Any]) -> SymbolicWhile:
+        assert dict_["kind"] == "while"
+        return SymbolicWhile(self._load_compound(dict_["body"]), self._load_value(dict_["condition"]))
+
+    def _load_if(self, dict_: t.Mapping[str, t.Any]) -> SymbolicIf:
+        assert dict_["kind"] == "if"
+        return SymbolicIf(
+            true_body=self._load_compound(dict_["trueBranchBody"]),
+            false_body=self._load_compound(dict_["falseBranchBody"]),
+            condition=self._load_value(dict_["condition"]))
```
ChrisTimperley, line 241: nitpick for style consistency: add a trailing comma to the condition line and put the closing bracket on its own line below

src/rosdiscover/recover/symbolic.py
```diff
@@ -349,6 +349,45 @@ def eval(self, context: SymbolicContext) -> None:
...
+
+    def to_dict(self) -> t.Dict[str, t.Any]:
+        return {
+            "kind": "while",
+            "body": [self.body.to_dict()],
+            "condition": [self.condition.to_dict()],
+        }
+
+    def eval(self, context: SymbolicContext) -> None:
+        self.condition.eval(context)
```
ChrisTimperley, line 387: This doesn't look quite right? Maybe we need to chat more about this one? Intuitively, I would have expected an actual while loop here.

src/rosdiscover/recover/symbolic.py
```diff
@@ -349,6 +349,45 @@ def eval(self, context: SymbolicContext) -> None:
...
+class SymbolicIf(SymbolicStatement):
+    """Represents a sequence of symbolic if."""
+    true_body: SymbolicCompound
+    false_body: SymbolicCompound
+    condition: SymbolicValue
+
+    def to_dict(self) -> t.Dict[str, t.Any]:
+        return {
+            "kind": "while",
+            "trueBranchBody": [self.true_body.to_dict()],
```
ChrisTimperley, line 362: same comment about wrapping with lists as below

src/rosdiscover/recover/symbolic.py
```diff
@@ -349,6 +349,45 @@ def eval(self, context: SymbolicContext) -> None:
...
+@attr.s(frozen=True, auto_attribs=True, slots=True)
+class SymbolicWhile(SymbolicStatement):
+    """Represents a sequence of symbolic while."""
+    body: SymbolicCompound
+    condition: SymbolicValue
+
+    def to_dict(self) -> t.Dict[str, t.Any]:
+        return {
+            "kind": "while",
+            "body": [self.body.to_dict()],
```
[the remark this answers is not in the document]
tobiasduerschmid: fixed

src/rosdiscover/recover/loader.py

src/rosdiscover/recover/symbolic.py
```diff
@@ -349,6 +349,45 @@ def eval(self, context: SymbolicContext) -> None:
...
+        return {
+            "kind": "while",
+            "trueBranchBody": [self.true_body.to_dict()],
+            "falseBranchBody": [self.false_body.to_dict()],
+            "condition": [self.condition.to_dict()],
+        }
+
+    def eval(self, context: SymbolicContext) -> None:
+        self.condition.eval(context)
+        self.true_body.eval(context)
```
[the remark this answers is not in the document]
tobiasduerschmid: fixed

src/rosdiscover/recover/symbolic.py
tobiasduerschmid: fixed

src/rosdiscover/recover/symbolic.py
tobiasduerschmid: I've changed it to an if for now. If the condition evaluates to true we would have an endless loop here, which sounds like a bad idea. I think for now we only need the statements inside the loop to be evaluated ones if the condition is satisfiable since this is what happened before as well. Once we actually want them to be evaluated multiple times we can think about a sound way to evaluate this

PR #260, opened 2022-07-08
Print control conditions

tobiasduerschmid:
This PR builds on the control conditions from CXX recover and prints them to the console after analysis.


src/rosdiscover/recover/analyzer.py
```diff
@@ -51,22 +52,22 @@ def subscriber_callbacks(cls, program: SymbolicProgram) -> t.Set[SymbolicFunctio
         return result
 
     @classmethod
-    def publish_calls(cls, program: SymbolicProgram) -> t.Set[Publish]:
-        result = set()
+    def publish_calls(cls, program: SymbolicProgram) -> t.List[Publish]:
+        result = list()
```
ChrisTimperley, line 56: prefer `[]`

src/rosdiscover/recover/analyzer.py
```diff
@@ -51,22 +52,22 @@ def subscriber_callbacks(cls, program: SymbolicProgram) -> t.Set[SymbolicFunctio
...
+                if isinstance(stmt, Publish) and stmt not in result:
+                    result.append(stmt)
 
         return result
 
     @classmethod
-    def publish_calls_in_sub_callback(cls, program: SymbolicProgram) -> t.Set[Publish]:
-        result = set()
+    def publish_calls_in_sub_callback(cls, program: SymbolicProgram) -> t.List[Publish]:
+        result = list()
```
ChrisTimperley, line 66: same as above

src/rosdiscover/recover/analyzer.py
ChrisTimperley: also, just curious: why the change? is it to preserve order?

PR #261, opened 2022-07-08
Do not analyze unknown callbacks

tobiasduerschmid:
Add support for unknown callbacks

### DIFF
--- a/src/rosdiscover/recover/analyzer.py
+++ b/src/rosdiscover/recover/analyzer.py
@@ -44,6 +44,8 @@ class SymbolicProgramAnalyzer:
     def subscriber_callbacks(cls, program: SymbolicProgram) -> t.Set[SymbolicFunction]:
         result = set()
         for sub in cls.subscribers(program):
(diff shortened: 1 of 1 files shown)


src/rosdiscover/recover/analyzer.py
```diff
@@ -44,6 +44,8 @@ def rate_sleeps(cls, program: SymbolicProgram) -> t.Set[RateSleep]:
     def subscriber_callbacks(cls, program: SymbolicProgram) -> t.Set[SymbolicFunction]:
         result = set()
         for sub in cls.subscribers(program):
+            if sub.callback_name == "unknown":
```
ChrisTimperley, line 47: Relying on sentinel values isn't great. Can the callback name be an actual `Unknown` rather than a string?

tobiasduerschmid: That would make the system much more complex because symbolic strings need a symbolic context to be evaluated. So replacing the string value with a symbolic string would make all other places that use it much more complicated

PR #264, opened 2022-08-15
Use cached properties in SymbolicProgramAnalyzer

tobiasduerschmid:


src/rosdiscover/recover/analyzer.py
```diff
@@ -8,6 +8,7 @@
 import typing as t
 
 import attr
+from functools import cached_property
```
ChrisTimperley, line 11: bit of a nitpick: this should go into the first group of imports (standard library)

ChrisTimperley (approved):
Very nice! Other than one small linting comment, this looks good to me 👍🏻

src/rosdiscover/recover/analyzer.py
tobiasduerschmid: fixed

PR #265, opened 2022-08-17
State machine recovery (state vars and assignments in subscriber callbacks)

tobiasduerschmid:
First iteration on recovering state transitions. 
This PR adds the detection of state variables and the detection of relationships between their assignments and subscriber callbacks to rosdiscover


src/rosdiscover/recover/analyzer.py
```diff
@@ -10,21 +10,53 @@
...
-    SymbolicWhile
+    SymbolicWhile,
 )
 
 from .call import Publish, RateSleep
 from .call import Subscriber
 
 
-@attr.s(auto_attribs=True, slots=True)
+@attr.s(auto_attribs=True)  # Can't use slots with cached_property
```
ChrisTimperley, line 26: Ah. That makes sense. Good to know!

src/rosdiscover/recover/loader.py
```diff
@@ -211,6 +219,8 @@ def _load_value(self, dict_: t.Mapping[str, t.Any]) -> SymbolicValue:
             return self._load_reads_param(dict_)
         elif kind == "reads-param-with-default":
             return self._load_reads_param_with_default(dict_)
+        elif kind == "varRef":
```
ChrisTimperley, line 222: code smell: all other kinds use a form of snakecase (with `-` instead of _`), but `varRef` switches to camelCase

src/rosdiscover/recover/symbolic.py
```diff
@@ -842,6 +922,34 @@ def build(cls, entrypoint: str, functions: t.Iterable[SymbolicFunction]) -> Symb
             name_to_function[entrypoint] = SymbolicFunction.empty(entrypoint)
         return SymbolicProgram(entrypoint, name_to_function)
 
+    def func_of_stmt(self, stmt: SymbolicStatement) -> SymbolicFunction:
+        for func in self.functions.values():
+            if stmt in func.body:
```
ChrisTimperley, line 927: will this work for functions that contain compound statements?

ChrisTimperley (commented):
Just a few comments/questions, but otherwise this looks good! 👍🏻

src/rosdiscover/recover/symbolic.py
tobiasduerschmid: good question. Probably not. I'll verify

src/rosdiscover/recover/loader.py

src/rosdiscover/recover/symbolic.py
tobiasduerschmid: Fixed

ChrisTimperley (approved):
LGTM 👍🏻

PR #267, opened 2022-08-23
Load initial values

tobiasduerschmid:


src/rosdiscover/recover/loader.py
```diff
@@ -107,6 +107,7 @@ def _load_var_ref(self, dict_: t.Mapping[str, t.Any]) -> SymbolicVariableReferen
         return SymbolicVariableReference(
             variable=dict_["qualified_name"],
             type_=type_,
+            initial_value=self._load_value(dict_["initialValue"]),
```
ChrisTimperley, line 110: we should probably stick to kebab case for consistency in the JSON format (i.e., `initialValue` -> `initial-value`)

ChrisTimperley:
Technical debt is noted in #54 (storing initial value in declref). This looks good for now!

PR #268, opened 2022-08-23
Collect initial values of state variables from main method

tobiasduerschmid:


src/rosdiscover/recover/states_analyzer.py
```diff
@@ -41,10 +43,28 @@ def state_vars(self) -> t.List[SymbolicVariableReference]:
...
+            if var.initial_value.is_unknown():
+                for assign in self.main_state_var_assigns:
+                    if assign.variable == var.variable:
+                        value = assign.value if isinstance(assign.value, SymbolicValue) else SymbolicUnknown()
+                        result.append(SymbolicVariableReference(
+                            variable=var.variable,
+                            type_=var.type_,
+                            initial_value=value
+                        ))
+                        continue
```
ChrisTimperley, line 59: `continue` is implicit here, no?

src/rosdiscover/recover/states_analyzer.py
```diff
@@ -41,10 +43,28 @@ def state_vars(self) -> t.List[SymbolicVariableReference]:
                         var_refs.append(expr)
         return var_refs
 
+    @cached_property
+    def state_vars(self) -> t.List[SymbolicVariableReference]:
+        result: t.List[SymbolicVariableReference] = []
+        for var in self._state_vars:
+            if var.initial_value.is_unknown():
```
ChrisTimperley, line 50: suggest: `if not var.initial_value.is_unknown():\ncontinue` to reduce nesting/complexity

src/rosdiscover/recover/loader.py
```diff
@@ -114,7 +114,7 @@ def _load_variable_reference(self, dict_: t.Mapping[str, t.Any]) -> SymbolicVari
         assert dict_["kind"] == "variable-reference"
         type_ = SymbolicValueType.from_name(dict_["type"])
         return SymbolicVariableReference(
-            variable=dict_["variable"],
+            variable=dict_["qualified_name"],
```
ChrisTimperley, line 117: we probably don't want to change anything now, but we should be careful about mixing `_` and `-` as word separators in the JSON output.

src/rosdiscover/recover/symbolic.py
```diff
@@ -801,9 +801,9 @@ def children(self) -> t.Set[SymbolicExpr]:
     def to_dict(self) -> t.Dict[str, t.Any]:
         return {
             "kind": "variable-reference",
-            "variable": self.variable,
+            "qualified_name": self.variable,
```
ChrisTimperley, line 804: same comment here

ChrisTimperley (approved):
Other than a nitpick about word separators in the JSON file, this looks good to me 👍🏻

